PR #4, opened 2020-03-02
refactor: improve swagger tests, cleanup

zoltanpalfi:
part of AAE-2024


activiti-cloud-audit-service/activiti-cloud-starter-audit/src/main/java/org/activiti/cloud/starter/audit/configuration/AuditSwaggerConfig.java
```diff
@@ -16,28 +16,41 @@
...
+            .license(String.format("© %s-%s %s. All rights reserved",
+                buildProperties.get("inceptionYear"),
+                buildProperties.get("year"),
+                buildProperties.get("organization.name")))
+            .termsOfServiceUrl(buildProperties.get("organization.url"))
+            .build();
+    }
 
     @Bean
-    @ConditionalOnMissingBean(name = "auditApiSelector")
```
erdemedeiros, line 32: Any specific reason to remove `@ConditionalOnMissingBean`? It usually gives the flexibility to override this bean if needed. BTW we'll need to make changes at this level when examples for audit and query get merged together.

activiti-cloud-modeling-service/activiti-cloud-starter-modeling/src/main/java/org/activiti/cloud/starter/modeling/configuration/ModelingSwaggerConfig.java
```diff
@@ -34,12 +35,17 @@
...
-                .build();
+            .title(String.format("%s ReST API", buildProperties.getName()))
+            .description(buildProperties.get("description"))
+            .version(buildProperties.getVersion())
+            .license(String.format("© %s-%s %s. All rights reserved",
+                buildProperties.get("inceptionYear"),
+                buildProperties.get("year"),
+                buildProperties.get("organization.name")))
+            .termsOfServiceUrl(buildProperties.get("organization.url"))
+            .build();
```
erdemedeiros: This code is the same for all the 4 starters. We could maybe move this somewhere inside `activiti-cloud-services-swagger` and keep only the bean instantiation here (if needed). What do you think?

erdemedeiros (changes requested):
See my comments below. LGTM otherwise.

activiti-cloud-audit-service/activiti-cloud-starter-audit/src/main/java/org/activiti/cloud/starter/audit/configuration/AuditSwaggerConfig.java
zoltanpalfi: Was inconsistent, that's why I removed it, but I have added it back now.

activiti-cloud-modeling-service/activiti-cloud-starter-modeling/src/main/java/org/activiti/cloud/starter/modeling/configuration/ModelingSwaggerConfig.java
zoltanpalfi: That's exactly what we are planning to do, but as part of a next pull request/story.

activiti-cloud-audit-service/activiti-cloud-starter-audit/src/main/java/org/activiti/cloud/starter/audit/configuration/AuditSwaggerConfig.java
erdemedeiros: It makes sense, but in this case, it's better to add the annotation in the beans where it is missing.

erdemedeiros (approved):

PR #6, opened 2020-03-06
AAE-2026 add @ConditionalOnMissingBean annotation in order to overrid…

jesty:
add @ConditionalOnMissingBean annotation in order to override the bean in Query Service, merged with Notification, Audit

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ env:
   global:
     - MAVEN_CLI_OPTS="--batch-mode --errors --fail-at-end --show-version -DinstallAtEnd=true -DdeployAtEnd=false -U"
     - MAVEN_TEST_OPTS=" -Dspring.main.banner-mode=off -Denforcer.skip"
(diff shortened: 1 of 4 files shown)


zoltanpalfi (approved):
lgtm

zoltanpalfi (changes requested):
Can you open a separate PR @robfrank for your changes and don't use an already approved one please?

robfrank:
@zoltanpalfi  without that fix, the build was red. How to manage it in a better way, if any?

zoltanpalfi:
> @zoltanpalfi without that fix, the build was red. How to manage it in a better way, if any?

In general, you should separate concerns. Thus, you won't have a PR (and then a git comment) saying you add @ContidionalOnMissing annotation in the code while it's actually also changes the parent reference, the module ordering in the pom and so on.

One solution is to have it as a separate pull request those changes. Another solution would be to change this PR description & final commit message to reflect actual changes. Or another one, you can just have someone else to approve it... 😛

zoltanpalfi (approved):

PR #19, opened 2020-03-20
[WIP]: Replace org.json and Gson with Jackson

amohammedalfresco:
for Activiti/Activiti#3155
- Initial commit for org.json replacement with jackson library

### DIFF
--- a/activiti-cloud-modeling-service/pom.xml
+++ b/activiti-cloud-modeling-service/pom.xml
@@ -28,7 +28,8 @@
     <activiti-cloud-acceptance-tests.version>${project.version}</activiti-cloud-acceptance-tests.version>
     <activiti-cloud-modeling-service.version>${project.version}</activiti-cloud-modeling-service.version>
     <everit-json-schema.version>1.12.1</everit-json-schema.version>
-    <json.version>20180813</json.version>
+    <jackson.version>2.10.2</jackson.version>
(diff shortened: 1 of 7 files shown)


amohammedalfresco:
I'm hitting a wall here due to `everit-json-schema` heavily using `org.json` in their module, and to my surprise I've found they too are performing a task to replace `org.json` with `jackson` in [PR#190](https://github.com/everit-org/json-schema/pull/190).

I believe it would be wise to may be wait for their latest version which already replace `org.json` to `jackson` and so in theory we have less worry to fix the problems in our code.

activiti-cloud-modeling-service/pom.xml
```diff
@@ -28,7 +28,8 @@
     <activiti-cloud-acceptance-tests.version>${project.version}</activiti-cloud-acceptance-tests.version>
     <activiti-cloud-modeling-service.version>${project.version}</activiti-cloud-modeling-service.version>
     <everit-json-schema.version>1.12.1</everit-json-schema.version>
-    <json.version>20180813</json.version>
+    <jackson.version>2.10.2</jackson.version>
```
mteodori, line 31: versions for jackson are not needed, they come from spring boot

mteodori:
closing as won't fix, the everit library has rejected the PR to move to Jackson

PR #20, opened 2020-03-23
Restore default phases for attach sources and sign jars (#3184)

erdemedeiros:
Changing the default phases is causing issues while signing artifacts (sources jars and some pom files are not signed)
Ref https://github.com/Activiti/Activiti/issues/3184


igdianov (approved):
LGTM

PR #23, opened 2020-03-25
Remove commons-io dependency in activiti-cloud-services-commons-io

nithinnambiar:
Remove apache commons-io library from activiti-cloud-services-commons-io.
https://github.com/Activiti/Activiti/issues/3156


activiti-cloud-service-common/activiti-cloud-services-commons-io/pom.xml
```diff
@@ -24,8 +24,9 @@
       <artifactId>spring-data-commons</artifactId>
     </dependency>
     <dependency>
-      <groupId>commons-io</groupId>
-      <artifactId>commons-io</artifactId>
+      <groupId>junit</groupId>
```
robfrank, line 27: I don't think this dependency is needed

robfrank (changes requested):

activiti-cloud-service-common/activiti-cloud-services-commons-io/pom.xml
nithinnambiar: The unit test wouldn't compile without junit dependency.

mteodori: that is because the test is using the JUnit 4 while it should have been using the JUnit 5 API, it should be fixed

PR #32, opened 2020-04-02
feat(AAE-1291): add support for CloudBpmnError propagation 

igdianov:
This PR adds Api and implementation support for propagating business errors from cloud counectors using CloudBpmnError exception.

image

The implemented error infrastructure will allow to use error propagation mechanism to trigger Bpmn errors from cloud connectors as follows:

1. Provide special exception class in Cloud Api, i.e. CloudBpmnError, to designate as base class for throwing business exceptions from Cloud Connectors.
2. Recognize in Connector Error Handler that the exception is an instance of CloudBpmnError.
3. Propagate CloudBpmnError via CloudIntegrationBpmnErrorEvent message to Rb
4. Receive and handle CloudIntegrationBpmnErrorEvent in Rb by mapping CloudBpmnError to Java runtime exception instance
5. Provide Api to throw BpmnError from inside execution context to trigger existing BpmnError handling logic in Activiti Engine.
6. Capture BpmnError application event as part of the execution and wrap it in CloudBpmnErrorEvent message
7. Send CloudBpmnErrorEvent mesage to Audit and Query
8. Receive and save CloudBpmnErrorEvent in Audit database
9. Receive and save CloudBpmnErrorEvent in Query database

Test Coverage:

- [x] Basic Bpmn Error Handling 

image

- [x] Catch Bpmn End Error event and terminate process

image

- [x] Catch Bpmn Error Subprocess Event 
image

- [x] Catch Bpmn Error Boundary Event

image

Part of https://github.com/Activiti/Activiti/issues/2712

### DIFF
--- a/activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-connectors/src/main/java/org/activiti/services/connectors/channel/ServiceTaskIntegrationErrorEventHandler.java
+++ b/activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-connectors/src/main/java/org/activiti/services/connectors/channel/ServiceTaskIntegrationErrorEventHandler.java
@@ -90,7 +90,7 @@ public class ServiceTaskIntegrationErrorEventHandler {
                         ", flow node id `" + clientId +
                         "`. The integration error for the integration context `" + integrationContext.getId() + "` is {}";
 
-                LOGGER.error(message, integrationError);
+                LOGGER.info(message, integrationError);
 
                 if(CloudBpmnError.class.getName().equals(errorClassName)) {
                     if(execution.getActivityId().equals(clientId)) {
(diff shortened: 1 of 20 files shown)


activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-connectors/src/main/java/org/activiti/services/connectors/channel/ServiceTaskIntegrationErrorEventHandler.java
```diff
@@ -72,20 +81,42 @@ public void receive(IntegrationError integrationError) {
...
+                String errorClassName = integrationError.getErrorClassName();
+                String message = "Received integration error '" + errorClassName + "' with execution id `" +
                         integrationContextEntity.getExecutionId() +
-                        ", flow node id `" + integrationContext.getClientId() +
+                        ", flow node id `" + clientId +
                         "`. The integration error for the integration context `" + integrationContext.getId() + "` is {}";
 
-                LOGGER.debug(message, integrationError);
-
+                LOGGER.error(message, integrationError);
```
erdemedeiros: I'm wondering if this should be logged as `error` or something else (`debug` as before) in the sense where the error actually occurred in the connector and here we are only handling it. What do you think?

erdemedeiros (commented):
One small thought. LGTM otherwise.

activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-connectors/src/main/java/org/activiti/services/connectors/channel/ServiceTaskIntegrationErrorEventHandler.java
igdianov: I think info log level should be better

erdemedeiros (approved):

PR #61, opened 2020-04-16
AAE-1098 Incorrect process version

nithinnambiar:
When a process is created from the modelling app two rest calls are made which results in version getting updated twice. To fix this issue updateModelContent has been modified to make the version increment to model optional. By default version won't be incremented, if version requestParam is set to 'true'  in the rest call then the model version will be incremented.


robfrank (approved):
LGTM

PR #63, opened 2020-04-17
AAE-2382 persist start date in process started event

jesty:
The start date in process_instances table is always null.

After a debug session I noticed that when we consume a process creation (ProcessCreatedEventHandler) we try to persist the startDate but this is null. After in the start process event (ProcessStartedEventHandler) the date is valorized, but it is ignored.

Going to the root cause of the problem, the event PROCESS_CREATED event in RB is produced without the startDate because the startTime is null in org.activiti.engine.runtime.ProcessInstance, the source event entity.

IMHO a fix could be to persist the start date in ProcessStartedEventHandler.

### DIFF
--- a/activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/query/events/handlers/ProcessStartedEventHandler.java
+++ b/activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/query/events/handlers/ProcessStartedEventHandler.java
@@ -16,9 +16,6 @@
 
 package org.activiti.cloud.services.query.events.handlers;
 
-import java.util.Date;
-import java.util.Optional;
-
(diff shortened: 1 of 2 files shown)


eromano (approved):

zoltanpalfi:
We are missing tests. Any reason for not having at least a unit test?

zoltanpalfi:
we could add an assertion in `ProcessStartedEventHandlerTest`, for example in `handleShouldUpdateProcessInstanceStatusToRunningAndUpdateInstanceName`

jesty:
> we could add an assertion in `ProcessStartedEventHandlerTest`, for example in `handleShouldUpdateProcessInstanceStatusToRunningAndUpdateInstanceName`

Thanks for the advice. I just added an assertion in `handleShouldUpdateProcessInstanceStatusToRunningAndUpdateInstanceName`.

jesty:
> We are missing tests. Any reason for not having at least a unit test?

No, it was Friday afternoon and it was late, this is why I did a draft PR :)

zoltanpalfi (approved):
lgtm

PR #78, opened 2020-04-27
removes properties definition as it collide with YAML files

mathieubodin:
See https://github.com/Activiti/Activiti/issues/3256


erdemedeiros (approved):
Good spot, @mathieubodin! Thank you for your contribution!

PR #94, opened 2020-05-13
[AAE-2456] - added validation on project import

VitoAlbano:


EquinoxBlack (approved):

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/service/ProjectServiceImpl.java
```diff
@@ -387,6 +397,12 @@ private boolean isModelContent(String modelName,
                 && modelTypeService.isContentXML(modelType));
     }
 
+    public List<ModelValidationError> validateProjectName(Project project) {
+        Optional<ProjectValidator> projectNameValidator = projectValidators.stream()
+            .filter(projectValidator -> projectValidator instanceof ProjectNameValidator).findFirst();
+        return projectNameValidator.get().validate(project, null).collect(Collectors.toList());
```
zoltanpalfi, line 403: Using get on optional *without* checking if present means that if for whatever reason will be empty, it will blow up the code.

Please add a safety net for that:
```
if (projectNameValidator.isEmpty()) {
          return Collections.emptyList();  
        }
```
Alternatively, refactor the stream operations so it does that

PR #95, opened 2020-05-14
Fixed license header to 2020

mteodori:
Part of Activiti/Activiti#3154


activiti-cloud-acceptance-tests/activiti-cloud-acceptance-tests-core/src/main/java/org/activiti/cloud/acc/core/assertions/ProcessAssertions.java
```diff
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2017-2020 Alfresco.
```
zoltanpalfi: in Activiti we have `Alfresco Software, Ltd.`, shouldn't it be the same here?

zoltanpalfi (commented):
one question, but that is replicated in the whole PR

activiti-cloud-acceptance-tests/activiti-cloud-acceptance-tests-core/src/main/java/org/activiti/cloud/acc/core/assertions/ProcessAssertions.java
mteodori: yes it should, let me check where that comes from

zoltanpalfi (approved):
Lgtm! 👍

zoltanpalfi (approved):

PR #111, opened 2020-05-29
Serialize BigDecimals as String

erdemedeiros:
This will ensure that BigDecimals are serialized as String and not as a number, meaning that double quotes will be added around the value. Serializing it as a number it's problematic because, by default, it will be deserialized back to Java as double and it will lose precision. For instance, `1.00` (scale 2) will become `1.0` (scale 1) that are considered as different values in BigDecimal. By adding the quotes, it will be deserialized as String, but it will not lose the information about the scale, so it can be easily converted back to BigDecimal.


erdemedeiros:
Ref https://github.com/Activiti/Activiti/issues/3215

igdianov (approved):
ltgm

PR #117, opened 2020-06-02
Support filtering tasks by variable name and value

erdemedeiros:
Allow filtering tasks by variable name and value on query service.

`v1/tasks?variables.name=firstName&variables.value=John`

By default `variables.value` will be parsed as a String. It's possible to explicitly define the by adding the extra parameter `variables.type`:

`v1/tasks?variables.name=approved&variables.value=true&variables.type=boolean`

Ref https://github.com/Activiti/Activiti/issues/3309


igdianov (approved):
LGTM

PR #113, opened 2020-06-01
[AAE-2673] Enable global models in database

pmartinezga:
Enable models to belong to more than one project by setting the current OneToMany relationship between project and models to ManyToMany

PR #118, opened 2020-06-02
[AAE-2679] Add new REST endpoints to handle global models

pmartinezga:
Add the necessary REST endpoints to manage the relationships between projects and models as now models can be globals. 

This PR depends on #113  to be merged in first place.

### DIFF
--- a/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/main/java/org/activiti/cloud/services/modeling/rest/api/ModelRestApi.java
+++ b/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/main/java/org/activiti/cloud/services/modeling/rest/api/ModelRestApi.java
@@ -312,4 +312,15 @@ public interface ModelRestApi {
         @ApiParam(value = DELETE_RELATIONSHIP_MODEL_PROJECT_MODEL_ID_PARAM_DESCR, required = true)
         @PathVariable String modelId);
 
+    @ApiOperation(
+        tags = MODELS,
+        value = "Create new model that does note belong to a project",
+        notes = "Create a new model with no relationship to other projects"
+    )
+    @PostMapping(path = "/models")
+    @ResponseStatus(CREATED)
+    EntityModel<Model> createModelWithoutProject(
+        @ApiParam(CREATE_MODEL_PARAM_DESCR)
+        @RequestBody Model model);
+
 }
--- a/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/main/java/org/activiti/cloud/services/modeling/rest/api/ModelRestApi.java
+++ b/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/main/java/org/activiti/cloud/services/modeling/rest/api/ModelRestApi.java
@@ -279,7 +279,7 @@ public interface ModelRestApi {
         @ApiParam(GET_MODELS_TYPE_PARAM_DESCR)
         @RequestParam(MODEL_TYPE_PARAM_NAME) String type,
         @ApiParam(value = INCLUDE_ORPHANS_PARAM_DESCR, required = false)
-        @RequestParam(value = INCLUDE_ORPHANS_PARAM_NAME, required = false) Boolean includeOrphans,
+        @RequestParam(value = INCLUDE_ORPHANS_PARAM_NAME, required = false, defaultValue = "false") boolean includeOrphans,
         Pageable pageable);
 
     @ApiOperation(
@@ -297,7 +297,7 @@ public interface ModelRestApi {
         @ApiParam(value = SCOPE_PARAM_DESCR, required = false)
         @RequestParam(value = SCOPE_PARAM_NAME, required = false) String scope,
         @ApiParam(value = FORCE_PARAM_DESCR, required = false)
-        @RequestParam(value = FORCE_PARAM_NAME, required = false) Boolean force);
+        @RequestParam(value = FORCE_PARAM_NAME, required = false, defaultValue = "false") boolean force);
 
     @ApiOperation(
         tags = MODELS,
--- a/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/main/java/org/activiti/cloud/services/modeling/rest/api/ModelRestApi.java
+++ b/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/main/java/org/activiti/cloud/services/modeling/rest/api/ModelRestApi.java
@@ -96,6 +96,26 @@ public interface ModelRestApi {
 
     String PROJECT_ID_PARAM_NAME = "projectId";
 
+    String INCLUDE_ORPHANS_PARAM_DESCR = "If true, then models with no relationship to any project are retrieved regardless of their scope";
+
+    String INCLUDE_ORPHANS_PARAM_NAME = "includeOrphans";
+
+    String RELATE_MODEL_PROJECT_PROJECT_ID_PARAM_DESCR = "The id of the project to associate the model with";
+
+    String RELATE_MODEL_PROJECT_MODEL_ID_PARAM_DESCR = "The id of the model to associate the project with";
+
+    String SCOPE_PARAM_DESCR = "Scope to update the model if needed (optional)";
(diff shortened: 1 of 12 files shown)


activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/main/java/org/activiti/cloud/services/modeling/rest/api/ModelRestApi.java
```diff
@@ -246,4 +266,61 @@ void validateModelExtensions(
...
+        notes = "Get the models that has GLOBAL as scope. " +
+            "Minimal information for each model is returned."
+        //response = AlfrescoModelPage.class
+    )
+    @GetMapping(path = "/models")
+    PagedModel<EntityModel<Model>> getGlobalModels(
+        @ApiParam(GET_MODELS_TYPE_PARAM_DESCR)
+        @RequestParam(MODEL_TYPE_PARAM_NAME) String type,
+        @ApiParam(value = INCLUDE_ORPHANS_PARAM_DESCR, required = false)
+        @RequestParam(value = INCLUDE_ORPHANS_PARAM_NAME, required = false) Boolean includeOrphans,
```
erdemedeiros: Here (and in the implementation), you should be able to use `boolean` instead of `Boolean` so you can save checks for null.

If this change causes NullPointerException while loading the parameter you can use ``defaultValue = "false"`` to set the default value.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/main/java/org/activiti/cloud/services/modeling/rest/api/ModelRestApi.java
```diff
@@ -246,4 +266,61 @@ void validateModelExtensions(
...
+        Pageable pageable);
+
+    @ApiOperation(
+        tags = MODELS,
+        value = "Add or update the relationship between an existing model, and the project",
+        notes = "Get the model associated with the project updated. " +
+            "Minimal information for the model is returned."
+    )
+    @PutMapping(path = "/projects/{projectId}/models/{modelId}")
```
erdemedeiros, line 291: Shouldn't this go inside `ProjectRestApi`, as the path starts with `/projects`?

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/main/java/org/activiti/cloud/services/modeling/rest/controller/ModelController.java
```diff
@@ -232,6 +233,71 @@ public void validateModelExtensions(
...
+        @RequestParam(value = SCOPE_PARAM_NAME, required = false) String scope,
+        @RequestParam(value = FORCE_PARAM_NAME, required = false) Boolean force) {
+
+        Project project = projectController.findProjectById(projectId);
+        Model modelToBeUpdated = findModelById(modelId);
+        Model newModel = findModelById(modelId);
+
+        if (force != null && force &&
+            (
+                (scope != null && scope.equals(ModelScope.PROJECT.name()))
```
erdemedeiros: You can flip `equals` method to avoid null check: `ModelScope.PROJECT.name().equals(scope)`

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/main/java/org/activiti/cloud/services/modeling/rest/controller/ModelController.java
```diff
@@ -232,6 +233,71 @@ public void validateModelExtensions(
...
+    @Override
+    public EntityModel<Model> putProjectModelRelationship(
+        @PathVariable String projectId,
+        @PathVariable String modelId,
+        @RequestParam(value = SCOPE_PARAM_NAME, required = false) String scope,
+        @RequestParam(value = FORCE_PARAM_NAME, required = false) Boolean force) {
+
+        Project project = projectController.findProjectById(projectId);
+        Model modelToBeUpdated = findModelById(modelId);
+        Model newModel = findModelById(modelId);
```
erdemedeiros: It sounds strange to me use `findModelById(modelId)` to initialize both `modelToBeUpdated` and `newModel`. It's probably reaching the cache, but otherwise, it would mean that we ha two access to the database to retrieve the same object at the same place. I was wondering if you cannot use the same variable in the update. I mean `modelService.updateModel(newModel, newModel)`.

BTW, as the controller is annotated with `@Transactional`, I think the only reason for that you need to call `modelService.updateModel` is because it has some validation logic. Otherwise, your changes will be probably be taken into account event without calling `updateModel`.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/main/java/org/activiti/cloud/services/modeling/rest/controller/ModelController.java
```diff
@@ -232,6 +233,71 @@ public void validateModelExtensions(
...
+    }
+
+    @Override
+    public EntityModel<Model> deleteProjectModelRelationship(
+        @PathVariable String projectId,
+        @PathVariable String modelId) {
+
+        Project project = projectController.findProjectById(projectId);
+        Model modelToBeUpdated = findModelById(modelId);
+        Model newModel = findModelById(modelId);
```
erdemedeiros: Same question here.

erdemedeiros (changes requested):
Here is my feedback.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/main/java/org/activiti/cloud/services/modeling/rest/api/ModelRestApi.java
pmartinezga: You're right, more clear for sure

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/main/java/org/activiti/cloud/services/modeling/rest/api/ModelRestApi.java
pmartinezga: Although the path points to `projects`, the service that is going to be called is the `ModelService` and that is why I put it here. I just tried to follow the example of what was here already for example in this line https://github.com/Activiti/activiti-cloud/blob/db266925733464f6465f5465a609e4200f1b7cfe/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/main/java/org/activiti/cloud/services/modeling/rest/api/ModelRestApi.java#L127

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/main/java/org/activiti/cloud/services/modeling/rest/controller/ModelController.java
pmartinezga: Regarding your 2nd concern, there are further validations in fact (https://github.com/Activiti/activiti-cloud/blob/db266925733464f6465f5465a609e4200f1b7cfe/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/service/ModelServiceImpl.java#L178)

Regarding first, you are right, in this case there is no need for having two different variables.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/main/java/org/activiti/cloud/services/modeling/rest/controller/ModelController.java
pmartinezga: Yes, same as before 👍

erdemedeiros (approved):
LGTM

PR #156, opened 2020-06-29
fix(AAE-2166); command context close listener transacted producer

igdianov:
I have investigated the issue with CommandContextCloseListener.closed being called inside transaction. By design, the close method is called at the end of Activiti engine transaction when CommandContext is successfully closed and no rollback has happened. However, 

As part of the code review, I have looked into whether the transmission of the message on the producer side is linked to the spring transaction or whether we need to enable a Spring Cloud Stream feature. The problem I discovered is that there could be a potential problem with publishing messages without having producer channel participating in the Spring transaction, which will result in events published even if Activiti transaction is rolled back. 

Spring Cloud Stream Rabbit binder implementation has transaction synchronization configuration with Spring PlatformTransactionManager for producer output channels

To enable transactional support, the producer output stream in rabbit binder configuration needs to be declared as transacted: spring.cloud.stream.rabbit.bindings.output.producer.transacted=true

After that, the MessageProducerCommandContextCloseListener.closed(CommandContext commandContext) method should be annotated with @Transactional to join Activiti transaction when sending events.

If Activiti successfully commits its transaction, the producer binder will also execute commit on transacted channel to publish messages.

In the event Activiti database transaction is rolled back, the producer binder will rollback any messages waiting for commit in the output stream Rabbit channel.

If there is infrastructure problem and messages cannot be sent, producer send() method will return false and raise ActivitiException to signal Activiti closing manager to rollback database transaction.


erdemedeiros:
Ref https://github.com/Activiti/Activiti/issues/1801

erdemedeiros (approved):

PR #160, opened 2020-07-08
Activiti/Activiti#3369 Upgrade to keycloak client to 10.0.2 

miguelruizdev:
Addressing: https://github.com/Activiti/Activiti/issues/3369 / https://github.com/Activiti/Activiti/issues/3358

Important notes:
There's a bug with this version, already fixed, but only available for `11.0.0`, as reflected in this ticket: https://issues.redhat.com/browse/KEYCLOAK-14520

The present PR is a workaround until `11.0.0` comes out

### DIFF
--- a/activiti-cloud-service-common/activiti-cloud-services-common-security-keycloak/src/main/java/org/activiti/cloud/services/common/security/keycloak/config/CustomKeycloakResolverConfiguration.java
+++ b/activiti-cloud-service-common/activiti-cloud-services-common-security-keycloak/src/main/java/org/activiti/cloud/services/common/security/keycloak/config/CustomKeycloakResolverConfiguration.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2017-2020 Alfresco Software, Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.activiti.cloud.services.common.security.keycloak.config;
 
 import org.keycloak.adapters.KeycloakConfigResolver;
--- a/activiti-cloud-service-common/activiti-cloud-services-common-security-keycloak/src/main/java/org/activiti/cloud/services/common/security/keycloak/config/CustomKeycloakResolverConfiguration.java
+++ b/activiti-cloud-service-common/activiti-cloud-services-common-security-keycloak/src/main/java/org/activiti/cloud/services/common/security/keycloak/config/CustomKeycloakResolverConfiguration.java
@@ -1,0 +1,15 @@
+package org.activiti.cloud.services.common.security.keycloak.config;
+
+import org.keycloak.adapters.KeycloakConfigResolver;
+import org.keycloak.representations.adapters.config.AdapterConfig;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class CustomKeycloakResolverConfiguration {
+
+    @Bean("keycloakConfigResolver")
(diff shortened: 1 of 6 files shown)


activiti-cloud-service-common/activiti-cloud-services-common-security-keycloak/src/main/java/org/activiti/cloud/services/common/security/keycloak/config/CustomKeycloakResolverConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package org.activiti.cloud.services.common.security.keycloak.config;
+
+import org.keycloak.adapters.KeycloakConfigResolver;
+import org.keycloak.representations.adapters.config.AdapterConfig;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class CustomKeycloakResolverConfiguration {
```
ffazzini, line 24: is it in a different class so it is easier to remove when the 11 version gets published?

miguelruizdev:
The **adapterConfig** of the configResolver was null, hence the NPE - the already mentioned bug in keycloak `10.0.2`

We need to create a custom keycloak config resolver, with the adapter config and inject it into CommonSecurityAutoConfiguration
For that, the bean declaration of the custom keycloak config resolver needs to be outside that class, and be created first to allow its injection

activiti-cloud-service-common/activiti-cloud-services-common-security-keycloak/src/main/java/org/activiti/cloud/services/common/security/keycloak/config/CustomKeycloakResolverConfiguration.java
miguelruizdev: exactly, when we update to 11 we only need to get rid of the class, the configuration where this bean is initialized and clean up the `spring.factories`

ffazzini: ok better to have an 11 issue pointing to this code already. Thanks

miguelruizdev: actually, as said here: https://github.com/Activiti/activiti-cloud/pull/160#issuecomment-655582491 , 
that config needs to be a different one, in charge of initializing the bean with the adapter configuration that the CommonSecurityAutoConfiguration uses at startup time

activiti-cloud-service-common/activiti-cloud-services-common-security-keycloak/src/main/java/org/activiti/cloud/services/common/security/keycloak/config/CustomKeycloakSpringConfigResolver.java
```diff
@@ -0,0 +1,45 @@
...
+    private KeycloakDeployment keycloakDeployment;
+
+    @Autowired
+    public CustomKeycloakSpringConfigResolver(AdapterConfig adapterConfig) {
+        this.adapterConfig = adapterConfig;
+    }
+
+    @Override
+    public KeycloakDeployment resolve(OIDCHttpFacade.Request request) {
+        if (keycloakDeployment != null) {
```
ffazzini, line 37: this does not look nice but seems to be the main point of the workaround suggested right?

ffazzini (approved):

activiti-cloud-service-common/activiti-cloud-services-common-security-keycloak/src/main/java/org/activiti/cloud/services/common/security/keycloak/config/CustomKeycloakSpringConfigResolver.java
miguelruizdev: actually, it's an exact copy of the existing class in keycloak: https://github.com/keycloak/keycloak/blob/master/adapters/oidc/spring-boot-adapter-core/src/main/java/org/keycloak/adapters/springboot/KeycloakSpringBootConfigResolver.java#L37
😄

mteodori (approved):

erdemedeiros (approved):
LGTM

PR #163, opened 2020-07-14
[AAE-2606] Add model updated listener registry and calls

pmartinezga:
There are some models such as files and scripts that need to update their extensions when the model information changes (in scripts and files, the uri must match the name of the model) so, as it is done in the content update, we are going to add a registry with model update listeners registered for each model type that needs them, that will be called just before persisting the model in the database when updating.


activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/test/java/org/activiti/cloud/services/modeling/rest/controller/GenericJsonModelTypeModelUpdateListenerControllerIT.java
```diff
@@ -0,0 +1,119 @@
...
+import org.springframework.test.web.servlet.MockMvc;
+import org.springframework.web.context.WebApplicationContext;
+
+/**
+ * Integration tests for models rest api dealing with Json models
+ */
+@ActiveProfiles(profiles = {"test", "generic"})
+@SpringBootTest(classes = ModelingRestApplication.class)
+@WebAppConfiguration
+@DirtiesContext(classMode = AFTER_EACH_TEST_METHOD)
```
erdemedeiros: Let's avoid `@DirtiesContext(classMode = AFTER_EACH_TEST_METHOD)` as it tends to make the test suite really slow. If the only purpose is to keep the database clear you can use `repository.deleteAll()` in the `@AfterEach`.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/service/ModelServiceImpl.java
```diff
@@ -179,6 +189,9 @@ public Model updateModel(Model modelToBeUpdated,
             newModel.getProjects().stream().forEach(project -> checkIfModelNameExistsInProject((Project) project,newModel));
         }
         checkModelScopeIntegrity(newModel);
+
+        emptyIfNull(findModelUpdateListeners(modelToBeUpdated.getType())).stream().forEach(listener -> listener.execute(modelToBeUpdated, newModel));
```
erdemedeiros: Instead of this `emptyIfNull` check here, I think it would be better to move it to `findModelUpdateListeners`, in such a way that `findModelUpdateListeners` never return `null`.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/service/ModelingServiceAutoConfiguration.java
```diff
@@ -73,13 +73,15 @@ public ModelService modelService(ModelRepository modelRepository,
                                      ModelContentService modelContentService,
                                      ModelExtensionsService modelExtensionsService,
                                      JsonConverter<Model> jsonConverter,
-                                     ProcessModelContentConverter processModelContentConverter) {
+                                     ProcessModelContentConverter processModelContentConverter,
+                                     Set<ModelUpdateListener> modelUpdateListeners) {
```
erdemedeiros, line 77: Is there any default `ModelUpdateListener`? If not you need to mark it as optional (` @Autowired(required = false)`) otherwise the application will fail to start.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/test/java/org/activiti/cloud/services/modeling/rest/controller/GenericJsonModelTypeModelUpdateListenerControllerIT.java
```diff
@@ -0,0 +1,119 @@
...
+
+    private static final String GENERIC_MODEL_NAME = "simple-model";
+
+    @BeforeEach
+    public void setUp() {
+        mockMvc = webAppContextSetup(context).build();
+    }
+
+    @Test
+    public void should_callJsonModelUpdateListener_when_updatingModelContent() throws Exception {
```
erdemedeiros: I'd merge the two tests of this class in a single one and merge the `verify` part (check that the listener for which the type matching is called and the one for which the type is not matching is not called). The name could be changed to something like `should_callUpdateListenerMatchingWithModelType_when_updatingModelContent`.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/test/java/org/activiti/cloud/services/modeling/rest/controller/GenericNonJsonModelTypeModelUpdateListenerControllerIT.java
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * Integration tests for models rest api dealing with Json models
+ */
+@ActiveProfiles(profiles = {"test", "generic"})
+@SpringBootTest(classes = ModelingRestApplication.class)
+@WebAppConfiguration
+@DirtiesContext(classMode = AFTER_EACH_TEST_METHOD)
+@WithMockModelerUser
+public class GenericNonJsonModelTypeModelUpdateListenerControllerIT {
```
erdemedeiros: From what I understand this test is doing the same thing as the previous one but inverting the model type. In this case, I think this test can be removed as this behaviour is already covered by the previous test.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/service/ModelServiceImpl.java
```diff
@@ -90,21 +93,28 @@
...
+
     @Autowired
     public ModelServiceImpl(ModelRepository modelRepository,
                             ModelTypeService modelTypeService,
                             ModelContentService modelContentService,
                             ModelExtensionsService modelExtensionsService,
                             JsonConverter<Model> jsonConverter,
-                            ProcessModelContentConverter processModelContentConverter) {
+                            ProcessModelContentConverter processModelContentConverter,
+                            Set<ModelUpdateListener> modelUpdateListeners) {
```
erdemedeiros, line 105: Wouldn't it be better to use `List` here instead of `Set` as the type used by `modelUpdateListenersMapByModelType ` is `List`?

erdemedeiros (changes requested):
Here is my feedback.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/test/java/org/activiti/cloud/services/modeling/rest/controller/GenericJsonModelTypeModelUpdateListenerControllerIT.java
pmartinezga: Ok, in fact, it was not needed as we are not changing anything in DB in these tests. Removed from both `ModelUpdate ListenerControllerIT`

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/service/ModelingServiceAutoConfiguration.java
pmartinezga: No, there is no ModelUpdateListener by default (at least at the community level).

I am not completely sure that it will fail on startup as this very same `Registry pattern` has been implemented in the same way with the `ContentUpdateListener` and it doesn't seem to be failing. In fact, spring creates automatically an empty set for this collection of beans implementing the `ModelUpdateListener`interface if none is present.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/test/java/org/activiti/cloud/services/modeling/rest/controller/GenericJsonModelTypeModelUpdateListenerControllerIT.java
pmartinezga: OK, you are right, both tests do the same call to the API so we can do both verifications in the same test

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/test/java/org/activiti/cloud/services/modeling/rest/controller/GenericNonJsonModelTypeModelUpdateListenerControllerIT.java
pmartinezga: You are right as usual. The other tests are differentiating between JSON and no JSON model types because there are some special behaviors that depend on this aspect, but for the `ModelUpdateListener` there are no differences.

I will rename the Test Class to not to be tight to the model type, and remove the second one.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/service/ModelServiceImpl.java
pmartinezga: I have changed the `modelUpdateListenersMapByModelType` to be a map of string, set as I prefer the `modelUpdateListeners`to be a set and avoid duplications in the registration of listeners.

pmartinezga: We need it to be a set to avoid duplications in the listener registering and for Spring context to create an empty set if no listeners are registered. 
This is the same implementation followed in `ContentUpdateListener` registry pattern.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/service/ModelingServiceAutoConfiguration.java
erdemedeiros: Ok, It sounds good then. I had some trouble with that in the past, but it looks like it has been fixed in the latest versions of Spring.

erdemedeiros (approved):
LGTM

PR #166, opened 2020-07-16
Fix travis configuration to stop on test failure

almerico:


erdemedeiros:
Fixes https://github.com/Activiti/Activiti/issues/3376

PR #168, opened 2020-07-17
Fix IntegrationRequestSender

erdemedeiros:
- the integration request itself is sent after the activiti transaction has been committed because the connector channel is not transacted, so sending it before commit could result in sending an integration request for a rolled back transaction
- the audit event related to integration request is included in the Activiti transaction because the audit channel is transacted, so the message will only be available to consumers if the the Activiti transaction has been successfully committed

Ref  Activiti/Activiti#1801


igdianov (approved):
LGTM

PR #172, opened 2020-07-23
fix: add call simple subProcess with expression integration test

igdianov:


erdemedeiros:
@igdianov, it looks good to me. However, after reflection, even if I initially suggested to add the test in the runtime bundle, it's probably better to have it directly in the starter of the core module as it's not involving any cloud-specific behavior. [Here](https://github.com/Activiti/Activiti/blob/develop/activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ProcessRuntimeCallActivityIT.java) for instance. What do you think?

igdianov:
> @igdianov, it looks good to me. However, after reflection, even if I initially suggested to add the test in the runtime bundle, it's probably better to have it directly in the starter of the core module as it's not involving any cloud-specific behavior. [Here](https://github.com/Activiti/Activiti/blob/develop/activiti-core/activiti-spring-boot-starter/src/test/java/org/activiti/spring/boot/process/ProcessRuntimeCallActivityIT.java) for instance. What do you think?

@erdemedeiros The test already exists in the core module. I think it makes sense to have cloud specific test because there is overriding cloud MappingAwareCallActivityBehaviour that should support call activity expressions.

erdemedeiros (approved):

PR #173, opened 2020-07-24
fix(AAE-3132): add call activity expression validator support

igdianov:
- [x] Implemented support for call activity variable reference validator
- [x] Implemented test coverage for call activity variable reference
- [x] Added missing test coverage for empty call activity
- [x] Added missing test coverage for valid call activity


erdemedeiros (approved):

erdemedeiros:
Ref https://github.com/Activiti/Activiti/issues/3410

PR #181, opened 2020-08-12
Filter task by process definition name

Giovanni007:
Issue: [#3398](https://github.com/Activiti/Activiti/issues/3398)


activiti-cloud-query-service/activiti-cloud-starter-query/src/main/resources/config/query/liquibase/master.xml
```diff
@@ -99,5 +99,25 @@
         splitStatements="true"
         stripComments="true"/>
   </changeSet>
+
+  <changeSet author="activiti-query"
+      id="alter5-schema-m9" dbms="postgresql">
+    <sqlFile dbms="postgresql"
+        encoding="utf8"
+        path="changelog/06-alter.pg.schema.7.1.0-m9.sql"
```
Giovanni007: What is the right name to use for the database scripts?

activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/query/events/handlers/TaskCreatedEventHandler.java
```diff
@@ -49,6 +48,7 @@ public void handle(CloudRuntimeEvent<?, ?> event) {
 
             
             queryTaskEntity.setProcessInstance(processInstanceEntity);
+            queryTaskEntity.setProcessDefinitionName(processInstanceEntity.getProcessDefinitionName());
         }
```
Giovanni007, line 53: The processDefinitionName is not present in the the internalTask entity in Activity. As temporary solution I take the processDefinitionName from the process instance. We need to discuss if it's worth adding that in Activity and if we can avoid duplications.

erdemedeiros: I'd prefer to keep this at the query service only, so we can keep the execution database as light as possible. So I think the current solution is a fair one.

activiti-cloud-query-service/activiti-cloud-starter-query/src/main/resources/config/query/liquibase/master.xml
erdemedeiros: The name sounds ok. The only point is that you need to be using `m10` suffix instead of `m9` as we are currently in the milestone 10. Also, this [another PR](https://github.com/Activiti/activiti-cloud/pull/194/files#diff-9644d6bb9ccce725c739db2cfcdd56f0R1) is also modifying the database it will have a conflicting name with yours, so better to go with `changelog/07-alter.pg.schema.7.1.0-m10.sql`

activiti-cloud-query-service/activiti-cloud-starter-query/src/main/resources/config/query/liquibase/master.xml
```diff
@@ -99,5 +99,25 @@
         splitStatements="true"
         stripComments="true"/>
   </changeSet>
+
+  <changeSet author="activiti-query"
+      id="alter5-schema-m9" dbms="postgresql">
```
erdemedeiros: `m10` suffix for the id (instead of `m9`).

activiti-cloud-query-service/activiti-cloud-starter-query/src/main/resources/config/query/liquibase/master.xml
```diff
@@ -99,5 +99,25 @@
...
+    <sqlFile dbms="postgresql"
+        encoding="utf8"
+        path="changelog/06-alter.pg.schema.7.1.0-m9.sql"
+        relativeToChangelogFile="true"
+        splitStatements="true"
+        stripComments="true"/>
+  </changeSet>
+
+  <changeSet author="activiti-query"
+      id="alter3-oracle-schema-m9" dbms="oracle">
```
erdemedeiros: Use `m10` suffix.

activiti-cloud-query-service/activiti-cloud-starter-query/src/main/resources/config/query/liquibase/master.xml
```diff
@@ -99,5 +99,25 @@
...
+    <sqlFile dbms="oracle"
+        encoding="utf8"
+        path="changelog/03-alter.oracle.schema.7.1.0-m9.sql"
```
erdemedeiros: Use `m10` suffix.

erdemedeiros (changes requested):
Some minor details.

erdemedeiros (approved):

PR #188, opened 2020-09-02
Activiti/Activiti#3418 Create client module and synchronize it with controller for process instances and definitions

miguelruizdev:
Addressing: Activiti/Activiti#3418


activiti-cloud-acceptance-tests/activiti-cloud-acceptance-tests-core/src/main/java/org/activiti/cloud/acc/core/helper/Filters.java
```diff
@@ -36,6 +36,16 @@
         return filteredCollection;
     }
 
+    public static Collection<CloudProcessInstance> checkProcessInstances(Collection<CloudProcessInstance> resource, String processKey) {
```
miguelruizdev, line 39: Needed a new method that accepts Collections instead of PagedModels.
The old method is still needed, since it's used by query and audit, services that still use manual clients

erdemedeiros (approved):
Great job! I really like the direction this taking!

PR #189, opened 2020-09-03
3422 - Added support for task due date filter

nithinnambiar:
https://github.com/Activiti/Activiti/issues/3422


activiti-cloud-query-service/activiti-cloud-starter-query/src/test/java/org/activiti/cloud/starter/tests/QueryTasksIT.java
```diff
@@ -1463,4 +1464,70 @@ public void shouldGetTaskUserCandidatesAfterTaskCompleted() {
...
+        sdf.setTimeZone(TimeZone.getTimeZone("GMT"));
+
+        Date dueDate = new Date();
+        Calendar cal = Calendar.getInstance();
+        cal.set(Calendar.SECOND, 0);
+        cal.set(Calendar.MILLISECOND, 0);
+        Date now = cal.getTime();
+
+        //set due date as current date + 1
+        dueDate.setTime(now.getTime() + 86400000);
```
erdemedeiros: Maybe you could use `Duration.ofDays(1).toMillis()` instead `86400000` as we can figure out that the duration is one day without any calculation.

activiti-cloud-query-service/activiti-cloud-starter-query/src/test/java/org/activiti/cloud/starter/tests/QueryTasksIT.java
```diff
@@ -1463,4 +1464,70 @@ public void shouldGetTaskUserCandidatesAfterTaskCompleted() {
...
+
+        Task assignedTask = taskEventContainedBuilder.anAssignedTaskWithDueDate("Assigned task",
```
erdemedeiros: Could you also include one task with the due date set to before and one with the due date set to after the desired dates to make sure that they will not be included in the result?

activiti-cloud-query-service/activiti-cloud-starter-query/src/test/java/org/activiti/cloud/starter/tests/QueryTasksIT.java
```diff
@@ -1463,4 +1464,70 @@ public void shouldGetTaskUserCandidatesAfterTaskCompleted() {
...
+        eventsAggregator.sendAll();
+
+        await().untilAsserted(() -> {
+            //when
+            //set check date to current date
+            Date fromDate = now;
+            // to date, from date plus 2 days
+            Date toDate = new Date(dueDate.getTime() + 86400000);
+            ResponseEntity<PagedModel<Task>> responseEntity = testRestTemplate
+                .exchange(TASKS_URL + "?dueDate=" + sdf.format(fromDate) + "&dueDate=" +
```
erdemedeiros: Here, I'm wondering if it's not confusing to have the same query parameter acting at the same time as equals or as a range. I mean someone seeing `?dueDate=2020-08-03&dueDate=2020-09-03` could think that this is acting as `in`. It's maybe clearer if we use separate parameters as we have for other dates (`dueDateFrom` and `dueDateTo` for range and `dueDate` for exactly match). What do you think?

erdemedeiros (changes requested):

activiti-cloud-query-service/activiti-cloud-starter-query/src/test/java/org/activiti/cloud/starter/tests/QueryTasksIT.java
nithinnambiar: I did initially tried to use that but I had to modify the TaskEntity and the Task schema which i thought wasn't the right thing to do. I am not that familiar with QueryDsl, do you know how to map different query parameters to the same attribute inside TaskRepository?

erdemedeiros (approved):
LGTM!

PR #192, opened 2020-09-04
feat(AAE-3197): Add support for display service task integration context 

igdianov:
This PR support for display service task integration context, i.e. status, error message, error stack trace, error date, etc.

- [x] GET /admin/v1/service-tasks/{serviceTaksId}/integration-context

### DIFF
--- a/activiti-cloud-api/activiti-cloud-api-process-model/src/main/java/org/activiti/cloud/api/process/model/CloudIntegrationContext.java
+++ b/activiti-cloud-api/activiti-cloud-api-process-model/src/main/java/org/activiti/cloud/api/process/model/CloudIntegrationContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2010-2020 Alfresco Software, Ltd.
+ * Copyright 2017-2020 Alfresco Software, Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


activiti-cloud-api/activiti-cloud-api-process-model/src/main/java/org/activiti/cloud/api/process/model/CloudIntegrationContext.java
```diff
@@ -0,0 +1,44 @@
...
+
+    Date getResultDate();
+
+    Date getErrorDate();
+
+    String getErrorMessage();
+
+    String getErrorClassName();
+
+    List<StackTraceElement> getStackTraceElements();
```
erdemedeiros, line 42: Open question: would it make sense to have these fields directly in `IntegrationContext` instead of `CloudIntegrationContext`?

erdemedeiros (commented):
I have an open question, but it looks good to me otherwise.

activiti-cloud-api/activiti-cloud-api-process-model/src/main/java/org/activiti/cloud/api/process/model/CloudIntegrationContext.java
igdianov: It will probably makes sense to extract api into IntegrationContext from CloudIntegrationContext. I can create PR in Activiti Core

erdemedeiros: As discussed together it does not add much value having this in the core after all.

erdemedeiros (approved):

erdemedeiros:
Ref https://github.com/Activiti/Activiti/issues/3426

PR #201, opened 2020-09-16
3437 filter tasks using completed by

nithinnambiar:
Added a new column in Task table to store 'completed by' value 
https://github.com/Activiti/Activiti/issues/3437

### DIFF
--- a/activiti-cloud-query-service/activiti-cloud-starter-query/src/main/resources/config/query/liquibase/changelog/06-alter.oracle.schema.7.1.0-m10.sql
+++ b/activiti-cloud-query-service/activiti-cloud-starter-query/src/main/resources/config/query/liquibase/changelog/06-alter.oracle.schema.7.1.0-m10.sql
@@ -1,5 +1,5 @@
 alter table task
-    add column completed_by varchar(255);
+    add completed_by varchar(255);
 
 create index task_completedBy_idx on task (completed_by);
 
--- a/activiti-cloud-query-service/activiti-cloud-starter-query/src/main/resources/config/query/liquibase/changelog/06-alter.oracle.schema.7.1.0-m10.sql
+++ b/activiti-cloud-query-service/activiti-cloud-starter-query/src/main/resources/config/query/liquibase/changelog/06-alter.oracle.schema.7.1.0-m10.sql
@@ -1,0 +1,6 @@
+alter table task
+    add column completed_by varchar(255);
+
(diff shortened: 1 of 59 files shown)


activiti-cloud-query-service/activiti-cloud-starter-query/src/main/resources/config/query/liquibase/changelog/06-alter.oracle.schema.7.1.0-m10.sql
```diff
@@ -0,0 +1,6 @@
+alter table task
+    add column completed_by varchar(255);
```
erdemedeiros: Syntax error for Oracle. It should be `add completed_by` instead of `add column completed_by`.

activiti-cloud-query-service/activiti-cloud-starter-query/src/test/java/org/activiti/cloud/starter/tests/QueryTasksIT.java
```diff
@@ -1583,10 +1584,51 @@ public void should_getTask_when_queryFilteredByProcessDefinitionName() {
...
+                    keycloakTokenProducer.entityWithAuthorizationHeader(),
+                    PAGED_TASKS_RESPONSE_TYPE
+                );
+
+            //then
+            assertThat(responseEntity).isNotNull();
+            assertThat(responseEntity.getStatusCode()).isEqualTo(HttpStatus.OK);
+            assertThat(responseEntity.getBody()).isNotNull();
+            Collection<Task> tasks = responseEntity.getBody().getContent();
+            assertThat(tasks).hasSize(1);
```
erdemedeiros: Note: `hasSize(1)` is redundant here as you're already using `containsExactly` (which will fail if the size is different from `1`).

erdemedeiros (changes requested):

activiti-cloud-query-service/activiti-cloud-starter-query/src/main/resources/config/query/liquibase/changelog/06-alter.oracle.schema.7.1.0-m10.sql
nithinnambiar: Good catch.

erdemedeiros (approved):

PR #203, opened 2020-09-18
 ACA-3708 - Added filter by initiators

Giovanni007:
Alloe user to filter process instances by initiator using a list of user

### DIFF
--- a/activiti-cloud-query-service/activiti-cloud-starter-query/src/test/java/org/activiti/cloud/starter/tests/QueryProcessInstancesEntityIT.java
+++ b/activiti-cloud-query-service/activiti-cloud-starter-query/src/test/java/org/activiti/cloud/starter/tests/QueryProcessInstancesEntityIT.java
@@ -565,4 +565,53 @@ public class QueryProcessInstancesEntityIT {
                                          PAGED_PROCESS_INSTANCE_RESPONSE_TYPE);
     }
 
+    @Test
+    public void shouldGetProcessInstancesFilteredByInitiator() {
+        
+        ProcessInstance processInstanceInitiatorUser1 = processInstanceBuilder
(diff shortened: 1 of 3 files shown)


activiti-cloud-query-service/activiti-cloud-starter-query/src/test/java/org/activiti/cloud/starter/tests/QueryProcessInstancesEntityIT.java
```diff
@@ -565,4 +565,53 @@ public void shouldGetProcessInstancesFilteredByCompletedDate() {
                                          PAGED_PROCESS_INSTANCE_RESPONSE_TYPE);
     }
 
+    @Test
+    public void shouldGetProcessInstancesFilteredByInitiator() {
```
ffazzini, line 569: why do we have two different assertions as part of this test?
The name of the test is quite generic `shouldGetProcessInstancesFilteredByInitiator` so i was expecting one call and one assertion.

Giovanni007: I will create a new method `shouldGetProcessInstancesFilteredByInitiatorList` for the second assertion

ffazzini: yes that can be an option, as second option if you have concerns on creating a new IT method because will slow down the build further.. you could make clearer in the test method why there are two different goals/assertions in the same method. First option sounds cleaner

erdemedeiros (approved):

erdemedeiros:
Ref https://github.com/Activiti/Activiti/issues/3462

PR #207, opened 2020-09-23
Create README.md

igdianov:


mteodori:
I think we should be more clear on when to use this module rather than the identity one (or use both and why) and update the identity README as well that looks outdated.

At the moment I see the security module included in any service including modeling:
* https://github.com/Activiti/activiti-cloud/search?q=activiti-cloud-services-common-security-keycloak

And identity in rb/query/audit/notifications:
* https://github.com/Activiti/activiti-cloud/search?q=activiti-cloud-services-common-identity-keycloak

mteodori:
we should also specify KeycloakAccessTokenPrincipalGroupsProvider requires extra configuration on keycloak to map groups as claims in the JWT

mteodori (approved):

PR #212, opened 2020-09-29
Make sure Maven Central is the first repository

erdemedeiros:
Most of the artifacts are coming from Maven Central, making sure it's the first in the list will speed up the build as it will avoid attempts to download artifacts from internal Nexus first.


erdemedeiros:
Outdated.

PR #216, opened 2020-10-05
Activiti/Activiti#3455 - Added jbehave FailingUponPendingStep to configuration

Giovanni007:
Configure Jbehave to fail when a step is pending.


erdemedeiros (approved):
LGMT

PR #218, opened 2020-10-07
fix(AAE-3761): Some BPMN errors are not shown in Service Task view

igdianov:
This PR fixes the problem with handling integration error events in Audit and Query that contain error message longer than 255 characters.

The fix will apply truncate function for error message entity attribute in order to avoid SQL error when saving error message contents to persistent backend.


mmartinadanx (approved):

erdemedeiros (approved):

erdemedeiros:
Ref https://github.com/Activiti/Activiti/issues/3426

PR #219, opened 2020-10-13
[AAE-3766] Set validator logging level to debug

pmartinezga:
When a user validates a model using the modeling-service, whenever the model is invalid the error and stack trace is printed in console because the logging level is set to ERROR. 

 

Change the logging level to DEBUG for all the validators loggers.


mmartinadanx (approved):

zoltanpalfi (approved):
Lgtm! 👍

PR #228, opened 2020-10-26
Feat(events) save previous value for update

zoltanpalfi:


activiti-cloud-audit-service/activiti-cloud-starter-audit/src/test/java/org/activiti/cloud/starter/audit/tests/it/AuditServiceIT.java
```diff
@@ -161,7 +161,7 @@ public void should_supportEventsContainingBigDecimalValues() {
         //given
```
erdemedeiros, line 161: I think you need to add an integration test here to make sure that `priviousValue` is stored in the database. I think it's currently ignored as I don't see any change to [VariableUpdatedEventEntity](https://github.com/Activiti/activiti-cloud/blob/develop/activiti-cloud-audit-service/activiti-cloud-services-audit-jpa/src/main/java/org/activiti/cloud/services/audit/jpa/events/VariableUpdatedEventEntity.java).

erdemedeiros (changes requested):

erdemedeiros (approved):

PR #230, opened 2020-11-02
Activiti/Activiti#3460- added event handler for application deployed

Giovanni007:
Requires [PR 3471]( https://github.com/Activiti/Activiti/pull/3471). Handle the application deployed event and create the Application Entity in query.


activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/query/events/handlers/ApplicationDeployedEventHandler.java
```diff
@@ -0,0 +1,75 @@
...
+        if(deployment.getName().equals(APPLICATION_DEPLOYED_EVENT_NAME)) {
+            ApplicationEntity application = new ApplicationEntity(
+                    deployment.getId(),
+                    applicationDeployedEvent.getAppName(),
+                    deployment.getVersion().toString()
+            );
+            
+            if(checkApplicationExist(application)) {
+                LOGGER.debug("Application " + application.getName() + " with version "
+                        + application.getName() + " already exists!");
```
andrea-ligios: `Name` is repeated two times, the second one in place of `version`.
Also, please let's replace concatenation with parameters templates: 

```
LOGGER.debug("Application {} with version {} already exists!",
                        application.getName(), application.getVersion());
```

activiti-cloud-query-service/activiti-cloud-starter-query/src/main/resources/config/query/liquibase/changelog/07-alter.oracle.schema.7.1.0-m11.sql
```diff
@@ -0,0 +1,13 @@
...
+    name                       varchar(255) not null,
+    version                    varchar(255),
+    primary key (id)
+);
+
+INSERT INTO application
+SELECT DISTINCT CONCAT(app_name, app_version), app_name, app_version
+FROM process_instance;
+
+commit;
```
andrea-ligios: Please add a newline

activiti-cloud-query-service/activiti-cloud-starter-query/src/main/resources/config/query/liquibase/changelog/10-alter.pg.schema.7.1.0-m11.sql
```diff
@@ -0,0 +1,11 @@
...
+(
+    id                         varchar(255) not null,
+    name                       varchar(255) not null,
+    version                    varchar(255),
+    primary key (id)
+);
+
+INSERT INTO application
+SELECT DISTINCT CONCAT(app_name, app_version), app_name, app_version
+FROM process_instance;
```
andrea-ligios: Please add a newline

andrea-ligios (changes requested):

andrea-ligios (approved):

activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/conf/EventHandlersAutoConfiguration.java
```diff
@@ -310,4 +312,11 @@ public IntegrationErrorReceivedEventHandler integrationErrorReceivedEventHandler
     public QueryEventHandlerContext queryEventHandlerContext(Set<QueryEventHandler> handlers) {
         return new QueryEventHandlerContext(handlers);
     }
+
+    @Bean
+    @ConditionalOnMissingBean
+    public ApplicationDeployedEventHandler applicationDeployedEventHandler(
+            ApplicationRepository processDefinitionRepository) {
```
erdemedeiros: Wrong naming for the parameter: it should be `applicationRepository`.

activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/query/events/handlers/ApplicationDeployedEventHandler.java
```diff
@@ -0,0 +1,75 @@
...
+    public ApplicationDeployedEventHandler(ApplicationRepository applicationRepository) {
+        this.applicationRepository = applicationRepository;
+    }
+
+    @Override
+    public void handle(CloudRuntimeEvent<?, ?> event) {
+        CloudApplicationDeployedEvent applicationDeployedEvent = (CloudApplicationDeployedEvent) event;
+        Deployment deployment = applicationDeployedEvent.getEntity();
+        LOGGER.debug("Handling application deployed event for " + deployment.getId());
+        if(deployment.getName().equals(APPLICATION_DEPLOYED_EVENT_NAME)) {
```
erdemedeiros: I think we can remove this check, as the runtime bundle is already filtering it and sending only events related to the deployment called `SpringAutoDeployment`.

activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/test/java/org/activiti/cloud/services/query/events/handlers/ApplicationDeployedEventHandlerTest.java
```diff
@@ -0,0 +1,110 @@
...
+        given(applicationRepository.exists(any())).willReturn(true);
+
+        //when
+        handler.handle(applicationDeployedFirstEvent);
+
+        //then
+        ArgumentCaptor<ApplicationEntity> applicationCaptor = ArgumentCaptor
+                .forClass(ApplicationEntity.class);
+
+        verify(applicationRepository, never()).save(applicationCaptor.capture());
```
erdemedeiros: I think it's better to use `any()` here instead o captor:
`verify(applicationRepository, never()).save(any());`

activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/test/java/org/activiti/cloud/services/query/events/handlers/ApplicationDeployedEventHandlerTest.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Test
+    public void getHandledEventShouldReturnProcessDeployedEvent() {
```
erdemedeiros: Wrong reference to ProcessDeployedEvent in the test name.

activiti-cloud-query-service/activiti-cloud-starter-query/src/main/resources/config/query/liquibase/changelog/07-alter.oracle.schema.7.1.0-m11.sql
```diff
@@ -0,0 +1,13 @@
...
+(
+    id                         varchar(255) not null,
+    name                       varchar(255) not null,
+    version                    varchar(255),
+    primary key (id)
+);
+
+INSERT INTO application
+SELECT DISTINCT CONCAT(app_name, app_version), app_name, app_version
+FROM process_instance;
```
erdemedeiros: I think it's better to use `process_definition` table as it also has `app_name` and `app_version`, but has fewer entries than `process_instance`. In addition, I'm wondering if this step is actually needed: with the current implementation, runtime bundle will send those events every time it starts, so basically this script will only be useful if we migrate query-service without migrating runtime bundle, what doesn't seem to be a real use case. What do you think?

erdemedeiros (changes requested):
Here is my feedback. In addition to my comments in the code, you also need to:
- add an entry for the new event in [CloudRuntimeEventType](https://github.com/Activiti/activiti-cloud/blob/develop/activiti-cloud-api/activiti-cloud-api-events/src/main/java/org/activiti/cloud/api/events/CloudRuntimeEventType.java) that's summarizing all the available events.
- Add or update and existing integration tests covering the new event in audit: https://github.com/Activiti/activiti-cloud/blob/develop/activiti-cloud-audit-service/activiti-cloud-starter-audit/src/test/java/org/activiti/cloud/starter/audit/tests/it/AuditServiceIT.java

erdemedeiros:
@Giovanni007, I don't see the new rest endpoint, are you adding it later on in another PR?

Giovanni007:
> @Giovanni007, I don't see the new rest endpoint, are you adding it later on in another PR?

Yes, I'll add it in a new PR .

activiti-cloud-query-service/activiti-cloud-starter-query/src/main/resources/config/query/liquibase/changelog/07-alter.oracle.schema.7.1.0-m11.sql
Giovanni007: I did that thinking a possible scenario where there are process instances saved before adding the handler for the event of application deployed in  Query. Happy to remove the script if this case will never happen in a real case. Thanks

erdemedeiros: There will be process instances already saved from previous versions, but on every start of the runtime bundle, it will send an ApplicationDeployedEvent for all deployments with the name `SpringAutoDepoyment`, even those that existed before. In the case where the runtime bundle starts before query service and sends the event before query is up and running, the message broker will keep the message and will deliver it to query when it starts listening to the events. So it should be safe to remove this... we can double-check by doing an application upgrade once it's merged.

erdemedeiros (approved):
LGTM!

PR #241, opened 2020-11-18
Changed resource descriptor name for the applications api response 

Giovanni007:
Right now when calling /applications with request mapping application/hal+json we get "applicationEntities" as name of the resources in the response. This PR is to fix the name and get "applications" as name of the list in the response. We will not use the request with hal+json but would be right change it to use the same naming convetions as other apis.

eg. current response
`{
    "_embedded": {
        "applicationEntities": [
            {`


andrea-ligios (approved):
LGTM!

PR #247, opened 2020-12-02
fix(GH-3574): Added root execution context message headers support 

### LINKED ISSUE #3574: Root execution context message headers support
so that we can inject root execution context attributes into engine events message headers to enable producer and consumers  inspect message headers and be able o correlate, route, and filter message events using the following root execution context headers if available:

- [x] rootProcessInstanceId
- [x] rootProcessDefinitionId
- [x] rootProcessDefinitionKey
- [x] rootProcessDefinitionVersion
- [x] deploymentId
- [x] deploymentName
- [x] deploymentVersion

Optional root execution context may include:

- [x] rootBusinessKey
- [x] rootProcessName
- [x] rootProcessDefinitionName

igdianov:
This PR adds support for injecting root execution context attributes into engine events message headers. The producer and consumers will be able inspect message headers to correlate, route, and filter message events with the following root exection context headers if available:

- [x] rootProcessInstanceId
- [x] rootProcessDefinitionId
- [x] rootProcessDefinitionKey
- [x] rootProcessDefinitionVersion
- [x] deploymentId
- [x] deploymentName
- [x] deploymentVersion

Optional root execution context may include:

- [x] rootBusinessKey
- [x] rootProcessName
- [x] rootProcessDefinitionName

This can be used to enable support for Spring Cloud Stream partitioned query and audit consumers with consistent message routing between partitions using rootProcessInstanceId hash value , i.e.

```
# split engine events stream to 10 partitions
spring.cloud.stream.bindings.auditProducer.producer.partition-count=10

# Use rootProcessInstanceId header to route message between partitions or use random hash value if missing header
spring.cloud.stream.bindings.auditProducer.producer.partition-key-expression=headers['rootProcessInstanceId']?:T(java.lang.Math).random()*100
```

Fixes https://github.com/Activiti/Activiti/issues/3574

### DIFF
--- a/activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-events/src/main/java/org/activiti/cloud/services/events/listeners/MessageProducerCommandContextCloseListener.java
+++ b/activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-events/src/main/java/org/activiti/cloud/services/events/listeners/MessageProducerCommandContextCloseListener.java
@@ -32,6 +32,7 @@ import org.springframework.util.Assert;
 @Transactional
 public class MessageProducerCommandContextCloseListener implements CommandContextCloseListener {
 
+    public static final String ROOT_EXECUTION_CONTEXT = "rootExecutionContext";
     public static final String PROCESS_ENGINE_EVENTS = "processEngineEvents";
 
     private final ProcessEngineChannels producer;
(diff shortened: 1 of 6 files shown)


erdemedeiros (changes requested):
@igdianov, as discussed earlier, it's probably better to make it explicitly in the header name that's the "root" execution. Also having an IT testing the headers with a Call Activity would be nice.

igdianov:
> @igdianov, as discussed earlier, it's probably better to make it explicitly in the header name that's the "root" execution. Also having an IT testing the headers with a Call Activity would be nice.

@erdemedeiros I have fixed the message header names and added test coverage for execution context message header values for execution with subprocesses.

erdemedeiros (approved):
LGTM!

PR #258, opened 2020-12-30
fix: intermittent test failures in messages hazelcast integration test

igdianov:
This PR improves Hazelcast test stability by changing tests to ITs for execution by maven-failsafe plugin.

Also, there is a minor change to set HZ group size to 3 in order to match HZ instance count.


erdemedeiros (approved):
Thanks you, @igdianov !

PR #259, opened 2020-12-30
fix: Override default test Hazelcast network configuration

igdianov:
This PR fixes HazelcastMessageStoreIT intermittent test failures due to Travis multi-cast networking configuration. The test will now use Tcp joiner config to avoid these problems.


erdemedeiros (approved):
Thank you, @igdianov !

PR #262, opened 2021-01-18
[Activiti/Activiti#3521] Create saveAs REST API

Nannakaroliina:
Addressing to issue: Activiti/Activiti#3521


activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/main/java/org/activiti/cloud/services/modeling/rest/api/ProjectRestApi.java
```diff
@@ -163,6 +165,23 @@ void exportProject(
                     required = false,
                     defaultValue = "true") boolean attachment) throws IOException;
 
+    @ApiOperation(
+            tags = PROJECTS,
+            value = "Save an project as zip file with chosen name",
+            notes = "This will create and download the zip with chosen name " +
+                    "containing the project folder and all related models.<br>")
+    @GetMapping(path = "/projects/{projectId}/saveAs")
```
zoltanpalfi: I don't think this is a GET, that's for reading an entity. That's more like a POST (or PUT if you think it really should be idempotent)

Naming convention is with dashes, so `/projects/{projectId}/save-as` in your case. I think naming also can be improved considering that this will actually create a new project (so it will have a new projectId as I understand it), maybe something like `copy` or `clone` represents better what we actually do

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/main/java/org/activiti/cloud/services/modeling/rest/api/ProjectRestApi.java
```diff
@@ -163,6 +165,23 @@ void exportProject(
...
+    void saveProjectAs(
+            HttpServletResponse response,
+            @ApiParam(value = EXPORT_PROJECT_ID_PARAM_DESCR, required = true)
```
zoltanpalfi: this is not export project call, if we re-use the same literal, let's rename it, otherwise we can create a new one

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/main/java/org/activiti/cloud/services/modeling/rest/api/ProjectRestApi.java
```diff
@@ -163,6 +165,23 @@ void exportProject(
...
+            @PathVariable String projectId,
+            @ApiParam(value = PROJECT_NAME_SAVEAS_DESCR)
+            @RequestParam(name = PROJECT_NAME_PARAM_NAME) String name,
+            @ApiParam(ATTACHMENT_API_PARAM_DESCR)
+            @RequestParam(name = EXPORT_AS_ATTACHMENT_PARAM_NAME,
```
zoltanpalfi: same, not, export project call, consider renaming constant or creating a new

Again, it's a bit strange to have it as attachment, but that overlaps with my previous comment on the http verb for this method

zoltanpalfi (changes requested):
I think we should revisit the design phase of how we want to implement this feature, currently we are using the wrong HTTP verb in my view

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/main/java/org/activiti/cloud/services/modeling/rest/api/ProjectRestApi.java
```diff
@@ -79,6 +81,8 @@
 
     String PROJECT_NAME_OVERRIDE_DESCR = "The name of the project that will override the current name of the project in the zip file";
 
+    String PROJECT_NAME_SAVEAS_DESCR = "The name of the project that will replace the original name of the project";
```
zoltanpalfi: Let's rename this to `PROJECT_NAME_COPY_DESCR` and state what we actually do, create a new project with this name, as what's currently there can be a bit misleading

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/test/java/org/activiti/cloud/services/modeling/rest/controller/ProjectControllerIT.java
```diff
@@ -858,4 +858,17 @@ public void should_throwBadRequestException_when_importingProjectWithInvalidName
             .andExpect(status().reason(is("Validation errors found in project's models")));
     }
 
+    @Test
+    public void should_returnZipFileWithNewProjectName_when_copyingProject() throws Exception {
```
zoltanpalfi: tests name needs to change to reflect changes, there's no longer a zip file returned

we could also test the case when we create a project with existing project-name

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/service/ProjectDescriptor.java
```diff
@@ -53,6 +53,8 @@ public String getName() {
         return project.getName();
     }
 
+    public void setName(String name) { this.project.setName(name); }
```
zoltanpalfi: format is off, everything is in one line, let's make it consistent

zoltanpalfi (commented):
few minor comment, but rest looks good

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/test/java/org/activiti/cloud/services/modeling/rest/controller/ProjectControllerIT.java
Nannakaroliina: Oh yes, didn't remember to fix the beginning of the name. I will add a second test as well

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/service/ProjectDescriptor.java
Nannakaroliina: this change is not needed anymore to be honest so I will just remove it

zoltanpalfi (approved):
Lgtm!

PR #322, opened 2021-02-10
fix(messages): add support for filtering non-serializable message headers

igdianov:
Kafka binder is now adding `kafka_consumer` message header with non-serializable Java object that breaks the serialization for the message events store aggregator. This PR adds support for messages service to be able sanitize Kafka headers from input channel and convert Spring Integration support channels injected into headers to string names.


mteodori (approved):
LGTM

erdemedeiros (approved):
LGTM!

PR #405, opened 2021-04-13
Refactor validators to re-use constructor

erdemedeiros:
Instead of using utility method from the internal interface, it's better to provide meaningful constructors


amohammedalfresco (approved):
LGTM!

PR #416, opened 2021-05-03
Create starter module including rabbit and kafka

erdemedeiros:
Create base module to include both rabbit and kafka Spring Cloud Stream binders. `rabbit` is the default one.
Ref https://github.com/Activiti/Activiti/issues/3625


igdianov (approved):
LGTM

PR #424, opened 2021-05-06
Disable rabbit health check when using kafka

erdemedeiros:
Otherwise rabbit health check will be triggered and make the application fail to start
Ref Activiti/Activiti#3625


igdianov (approved):
LGTM

PR #431, opened 2021-05-11
fix: Simplify engine events notifications handling using Spring Integration adapter to reactive stream

igdianov:
This PR simplifies engine events notifications processor and adds minor improvements;

 - [x]  Use Spring Integration adapter for reactive streams. 
- [x]  Remove complexity inherited from project Reactor  Sinks implementations causing message events subscription synchronization  issues.
- [x] Set consumer concurrency to 1 to ensure correct message ordering
- [x] Increase consumer prefetch to 20 for better performance


erdemedeiros (approved):
LGTM!

PR #437, opened 2021-05-31
test: reduce amount of logs produced by runtime bundle tests

erdemedeiros:
The file `ParsingError.bpmn` was added in order to test `NeverFailDeploymentStrategy`, making sure that even when an error occurs while parsing a bpmn file the application starts (ignoring the file that could not be parsed). However, as this file was added at the same folder as all the other processes, it's producing a stack trace every time one test starts the application. This change is moving `ParsingError.bpmn` to specific folder in a such way that the stack trace will be produced only once.


igdianov (approved):
LGTM

PR #438, opened 2021-06-01
Standardise Event Model: add capability to store json schemas in connector features

rbarroso:


pmartinezga (commented):
I was wondering if there is any chance of validating that the JSON schema is valid on the java side, have you investigated it?

pmartinezga (approved):
Forget my previous comment, maybe we use no valid JSON schemas in content model events as we will be extending cloud events as reference and the reference will not be present in the JSON schema.

So, all good for me!

erdemedeiros (approved):
LGTM!

erdemedeiros:
Ref https://github.com/Activiti/Activiti/issues/3670.

PR #444, opened 2021-06-23
fix(GH-3674):  split ProcessDeployedEvents into message per process definition

igdianov:
Part of https://github.com/Activiti/Activiti/issues/3674

This PR improves default strategy for publishing process deployed events using configured chunk size in runtime bundle properties. The default chunk size is set to 100 events per message.

To change the defaul chunk size, use `activiti.cloud.runtime-bundle.events-properties.chunk-size` configuration property.


erdemedeiros (approved):
LGTM!

PR #455, opened 2021-07-15
Upgrade to Spring Boot 2.5.3 and Spring Cloud 2020.0.3

gicappa:
Bumping a new **spring-boot/cloud** version

Found some _minor incompatibilities_ due to the evolution of the spring-boot/data platform from the **2.4.x** to the **2.5.x**

- The `Pageable` interface has a new method `withPage()` that has been implemented by delegating to the inner peageable delegate object implementation.
- `QueryRepositoryConfig` and `RepositoryRestConfig` are inheriting from `RepositoryRestConfigurer` that changed the signature of the method `configureRepositoryRestConfiguration()` adding the `CorsRegistry` parameter. The signature has been updated but the parameter has not been taken into consideration.
- In `ModelingRestExceptionHandler` the `DefaultErrorAttributes` class changed the signature of `getErrorAttributes()` switching from a boolean param to include the stack trace to a class `ErrorAttributeOptions` to manage several kind of error attributes to be included.

### DIFF
--- a/activiti-cloud-runtime-bundle-service/pom.xml
+++ b/activiti-cloud-runtime-bundle-service/pom.xml
@@ -8,13 +8,11 @@
     <version>7.1.0-SNAPSHOT</version>
     <relativePath>../activiti-cloud-build</relativePath>
   </parent>
-  <groupId>org.activiti.cloud</groupId>
   <artifactId>activiti-cloud-runtime-bundle-service-parent</artifactId>
   <version>7.1.0-SNAPSHOT</version>
   <packaging>pom</packaging>
   <name>Activiti Cloud Runtime Bundle Service :: Parent</name>
   <modules>
-
     <module>activiti-cloud-runtime-bundle-dependencies</module>
     <module>activiti-cloud-services-runtime-bundle</module>
     <module>activiti-cloud-starter-runtime-bundle</module>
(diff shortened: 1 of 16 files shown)


activiti-cloud-runtime-bundle-service/pom.xml
```diff
@@ -8,23 +8,42 @@
...
+https://github.com/spring-projects/spring-integration/commit/76f77ccd2c5f94e000495357e66c4f99fffc6010
+
+Since the issue happens only due to the dependencies imported for testing it has been fixed rolling
+back the library spring-integration-amqp the the previous version (5.5.1)
+-->
+      <dependency>
+        <groupId>org.springframework.integration</groupId>
+        <artifactId>spring-integration-amqp</artifactId>
+        <version>5.5.1</version>
+      </dependency>
```
erdemedeiros: You should be able to remove this, as it's already declared in `activiti-cloud-build-dependencies-parent`.

activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/main/java/org/activiti/cloud/alfresco/argument/resolver/AlfrescoPageRequest.java
```diff
@@ -82,6 +82,11 @@ public AlfrescoPageRequest first() {
                                        pageSize, getPageable());
     }
 
+    @Override
+    public Pageable withPage(int pageNumber) {
+        return pageable.withPage(pageNumber);
```
erdemedeiros: Actually, after double-checking, what you need to set here is the `skipCount` with `pageNumber * getPageSize()` because `getPageNumber` is overridden [here](https://github.com/Activiti/activiti-cloud/pull/455/files#diff-11ec11e53366eb734a751267e5ae0caea469ef247150dcab759642f9b2c891a1R51). Please, add a new test to  `AlfrescoPageRequestTest` to cover this.

erdemedeiros (changes requested):
Two small details. LGTM otherwise.

activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/main/java/org/activiti/cloud/alfresco/argument/resolver/AlfrescoPageRequest.java
gicappa: Fixed as you suggested, thanks.

activiti-cloud-runtime-bundle-service/pom.xml
gicappa: My first implementation was relying on the definition declared in `activiti-cloud-build-dependencies-parent` but `activiti-cloud-runtime-bundle-service` was failing again. I tried to inspect on the possible dependency issue but the complexity of the dependencies where preventing me to do it in a proper time and was diverting me from the primary goal of the pull request.

gicappa: Fixed by the commit 559fbc1

erdemedeiros (approved):
LGTM!

erdemedeiros:
Ref https://github.com/Activiti/Activiti/issues/3688

PR #462, opened 2021-08-11
Fix issue guava-18.0.jar

rbarroso:
Related issue https://github.com/Activiti/Activiti/issues/3701

### DIFF
--- a/activiti-cloud-notifications-graphql-service/dependencies/pom.xml
+++ b/activiti-cloud-notifications-graphql-service/dependencies/pom.xml
@@ -10,6 +10,9 @@
   <name>Activiti Cloud Notifications :: GraphQL Dependencies (Bill Of Materials)</name>
   <artifactId>activiti-cloud-notifications-graphql-dependencies</artifactId>
   <packaging>pom</packaging>
+  <properties>
+    <querydsl.version>5.0.0</querydsl.version>
+  </properties>
   <dependencyManagement>
     <dependencies>
       <dependency>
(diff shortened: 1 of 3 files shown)


miguelruizdev (approved):

activiti-cloud-notifications-graphql-service/dependencies/pom.xml
```diff
@@ -10,6 +10,9 @@
   <name>Activiti Cloud Notifications :: GraphQL Dependencies (Bill Of Materials)</name>
   <artifactId>activiti-cloud-notifications-graphql-dependencies</artifactId>
   <packaging>pom</packaging>
+  <properties>
+    <querydsl.version>5.0.0</querydsl.version>
+  </properties>
```
erdemedeiros, line 15: It's better to centralize the changes in a single place if it's possible. Have you tried to add the changes to [activiti-cloud-service-common-dependencies](https://github.com/Activiti/activiti-cloud/blob/develop/activiti-cloud-service-common/activiti-cloud-service-common-dependencies/pom.xml)? It should update it for both audit, query and notifications.

erdemedeiros (changes requested):
It's better to centralize the changes if it's possible.

activiti-cloud-notifications-graphql-service/dependencies/pom.xml
rbarroso: I totally agree: it's way better to centralise this changes. However, activiti-cloud-service-common-dependencies it's not  in the inheritance path of query, audit and notifications projects. Accordingly, we'd not update the querydsl version.
I've tested that with the dependency:tree.
Should we re-think that inheritance structure?

erdemedeiros: Ok, in this case, it's better to keep it like that. Re-thinking the inheritance structure would take time, and this is a temporary workaround as we should be able to remove it once spring boot references the new version of querydsl.

erdemedeiros (approved):
LGTM!

PR #481, opened 2021-09-09
fix(GH-3457): remove common security keycloak dependency from connectors

igdianov:
This PR removes the activiti-cloud-services-common-security-keycloak dependency from connector starter pom. This should resolved leaking activiti.keycloak.client properties and Keycloak client  beans into Connector applications.


erdemedeiros (approved):

erdemedeiros:
Ref https://github.com/Activiti/Activiti/issues/3457

PR #494, opened 2021-10-21
AAE-5950 migrate to OAS3

Giovanni007:
Ref: [3743](https://github.com/Activiti/Activiti/issues/3743) Update of the Springfox configuration and  swagger-core annotation to support OAS3

OAS3 doesn't support  '<< ' in the module names so  replacing them using "Of"
OAS3.0.3 doesn't support request body  for DELETE method ( eg. rb specifications will show an error  generates an error if trying to validate it on swagger-editor) but the support will be back with OAS3.1 (see [PR]( https://github.com/OAI/OpenAPI-Specification/pull/2117))


miguelruizdev (approved):
LGTM

balsarori (approved):
LGTM

PR #516, opened 2021-11-25
#3765 Add new Category field to process extensions

miguelruizdev:
Ref: https://github.com/Activiti/Activiti/issues/3765
Since from the Activiti engine we will leverage `targetNamespace` of definitions as process categories, we need to make sure that no model is without its category.
The default from modeling app (FE) can be an empty string, instead of the current broken link: `"http://bpmn.io/schema/bpmn"`

### DIFF
--- a/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/process/BpmnModelValidator.java
+++ b/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/process/BpmnModelValidator.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2017-2020 Alfresco Software, Ltd.
+ *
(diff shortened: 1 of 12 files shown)


activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/process/BpmnModelValidator.java
```diff
@@ -15,28 +15,32 @@
...
- * Interface for validating {@link BpmnModel} objects
- */
-public interface BpmnModelValidator extends ModelValidationErrorProducer {
+public class BpmnModelValidator implements BpmnCommonModelValidator {
+
+    public final String ERROR_TYPE = "Missing process category";
 
-    Stream<ModelValidationError> validate(BpmnModel bpmnModel,
-                                          ValidationContext validationContext);
+    public final String ERROR_DESCRIPTION = "The process category needs to be set";
```
jesty, line 29: Can you use a meaningful name? Thanks

jesty (approved):
LGTM, just a minor comment

PR #520, opened 2021-12-02
#3770 fix validate models and models extensions in order to generate a valid Swagger definition

jesty:
Ref: https://github.com/Activiti/Activiti/issues/3770


miguelruizdev (approved):
LGTM!

PR #544, opened 2022-01-05
fix: make sure validation also checks inner element containers

### LINKED ISSUE #3794: [Modeling] User task validations are not working inside subprocess
The validation done by the modeling application is not checking elements inside subprocesses.

Steps to reproduce
- create a process containing a subprocess
- inside the subprocess create a user task and leave it without an assignment
- click on the validation button

No validation error is produced while an error is expected as the user task has no assignment.

erdemedeiros:
In addition to direct flow elements, the validation should also check flow elements inside inner containers (subprocesses, event-subprocesses)

Fixes https://github.com/Activiti/Activiti/issues/3794


igdianov (approved):
LGTM

PR #545, opened 2022-01-06
fix: split m17 query liquibase Oracle DDL into separate scripts

igdianov:
This PR fixes query liquibase script errors for Oracle when creating sequences for process and task variables. I have split scripts into separate files to be able to specify statement execution policy per each script.


erdemedeiros (approved):
LGTM!

PR #577, opened 2022-02-17
Activiti/Activiti#3831 - Create juel simulator starter

nazarethjim:


pmartinezga (approved):
I think we should activate the Juel Expression Simulator in the modeling service by adding the `@EnableActivitiJuel` annotation in the class https://github.com/Activiti/activiti-cloud-application/blob/develop/activiti-cloud-modeling/src/main/java/org/activiti/cloud/modeling/ModelingApplication.java (as part as another PR, of course) to enable the simulator in the community version, because it is a feature that is going to be enabled in the modeling application ce.
What do you think @mmartinadanx

mmartinadanx (approved):

PR #581, opened 2022-02-18
AAE-7221 replace keycloak specific configuration with spring security configuration

jesty:
The PR remove the authorizations definition using Keycloak, replacing them with Spring Security.
In this PR there is refactoring in order to split a test rest template interceptor and the token producer. Moreover, the token has been abstracted. I did this refactoring in order to simplify the test in other modules like alfresco-process.


miguelruizdev (approved):

miguelruizdev:
Shouldn't this file also be refactored? : https://github.com/Activiti/activiti-cloud/blob/develop/activiti-cloud-runtime-bundle-service/activiti-cloud-starter-runtime-bundle/src/test/resources/access-control.properties

miguelruizdev (approved):
LGTM, I really like the step towards abstraction thanks to the new `IdentityTokenProducer`

PR #583, opened 2022-02-22
Activiti/Activiti#3834 Centralize keycloak configuration into security module

miguelruizdev:
The following properties:
`activiti.keycloak.client-id`
`activiti.keycloak.client-secret`
have been centralized into the existing keycloak security module.

The following ones:
`activiti.keycloak.test-user=testuser`
`activiti.keycloak.test-password=password`
have been renamed to a more abstract, non-keycloak related, name, since they are used only internally


jesty (approved):
LGTM

PR #587, opened 2022-02-24
limit process instances access to initiator

balsarori:
Ref: https://github.com/Activiti/Activiti/issues/3838

### DIFF
--- a/activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/query/rest/ProcessInstanceController.java
+++ b/activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/query/rest/ProcessInstanceController.java
@@ -110,10 +110,9 @@ public class ProcessInstanceController {
                                                                             processInstanceId,
                                                                             "Unable to find process instance for the given id:'" + processInstanceId + "'");
 
-        if (!securityPoliciesApplicationService.canRead(processInstanceEntity.getProcessDefinitionKey(),
-                                                        processInstanceEntity.getServiceName())) {
-            LOGGER.debug("User " + securityManager.getAuthenticatedUserId() + " not permitted to access definition " + processInstanceEntity.getProcessDefinitionKey());
-            throw new ActivitiForbiddenException("Operation not permitted for " + processInstanceEntity.getProcessDefinitionKey());
+        if (!canRead(processInstanceEntity)) {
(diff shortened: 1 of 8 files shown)


activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/security/ProcessInstanceRestrictionService.java
```diff
@@ -27,24 +30,40 @@
...
-        return restrictionBuilder.applyProcessDefinitionKeyFilter(predicate,
+        return restrictionBuilder.applyProcessDefinitionKeyFilter(initiatorPredicate,
                                                  securityPolicyAccess,
                                                  processInstanceFilter);
     }
 
+    private Predicate applyInitiatorRestriction(Predicate predicate) {
+        String userId = securityManager.getAuthenticatedUserId();
+
+        if (userId == null) {
```
balsarori, line 61: I'm not sure if there could be a scenario where the current user is null. I followed the same check here --> https://github.com/Activiti/activiti-cloud/blob/develop/activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/security/TaskLookupRestrictionService.java#L102

Giovanni007 (approved):
LGTM!

activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/query/rest/ProcessInstanceController.java
```diff
@@ -149,4 +147,10 @@ public ProcessInstanceController(ProcessInstanceRepository processInstanceReposi
                                                                                     pageable),
                                                   processInstanceRepresentationModelAssembler);
     }
+
+    private boolean canRead(ProcessInstanceEntity processInstanceEntity) {
```
mauriziovitale, line 151: what I don't like is that the canRead is not giving any clue about the initiator

mauriziovitale (approved):

activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/query/rest/ProcessInstanceController.java
balsarori: user needs to be the process initiator and have permissions to read the app `securityPoliciesApplicationService.canRead`

PR #619, opened 2022-03-02
task involved users can access process instance

balsarori:
Ref: https://github.com/Activiti/Activiti/issues/3850

### DIFF
--- a/activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/query/rest/ProcessInstanceController.java
+++ b/activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/query/rest/ProcessInstanceController.java
@@ -15,6 +15,7 @@
  */
 package org.activiti.cloud.services.query.rest;
 
+import java.util.List;
 import java.util.Optional;
 
 import org.activiti.api.runtime.shared.security.SecurityManager;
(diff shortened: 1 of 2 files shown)


activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-repo/src/main/java/org/activiti/cloud/services/query/app/repository/CustomizedTaskRepository.java
```diff
@@ -30,9 +32,11 @@
...
     Iterable<TaskEntity> findInProcessInstanceScope(Predicate predicate);
-    
-    Page<TaskEntity> findInProcessInstanceScope(Predicate predicate, 
+
+    Page<TaskEntity> findInProcessInstanceScope(Predicate predicate,
                                                 Pageable pageable);
 
     boolean existsInProcessInstanceScope(Predicate predicate);
+
+    boolean isUserInvolvedInProcessInstanceTasks(String processInstanceId, String userId, List<String> userGroups);
```
Giovanni007: Do you really need to add a custom method?  I think you can use the method provided by `TaskRepository`  as `exists` or `count` passing your custom predicate. Also, we have a set of Tasks in [ProcessInstanceEntity](https://github.com/Activiti/activiti-cloud/blob/8b6584cf36c1efb6170fc3f8240b3216bf1c0028/activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-model/src/main/java/org/activiti/cloud/services/query/model/ProcessInstanceEntity.java#L126). Can they be returned to avoid executing a new database call?

activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/query/rest/ProcessInstanceController.java
```diff
@@ -151,6 +158,13 @@ public ProcessInstanceController(ProcessInstanceRepository processInstanceReposi
     private boolean canRead(ProcessInstanceEntity processInstanceEntity) {
```
Giovanni007, line 158: is this not executed for the findAll method? what do we return in that case?

Giovanni007 (changes requested):
See comments

activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-repo/src/main/java/org/activiti/cloud/services/query/app/repository/CustomizedTaskRepository.java
balsarori: I tried to avoid adding query logic inside the ProcessInstanceController class. Also, `existsInProcessInstanceScope` will do a left join that I would like to avoid.

Giovanni007: I mean the methods provided by Querydls, not `existsInProcessInstanceScope`

mauriziovitale (approved):

Giovanni007 (approved):

PR #626, opened 2022-03-04
Fix Activiti 22.2.0-alpha.8 propagation

mteodori:
Fix the propagation from https://github.com/Activiti/Activiti/commits/22.2.0-alpha.8 which is not building

See: https://github.com/Activiti/Activiti/compare/22.2.0-alpha.7...22.2.0-alpha.8

Fixing test failures require exposing an existing functionality by adding an endpoint for getting variables as an admin user.

### DIFF
--- a/activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-rest-api/src/main/java/org/activiti/cloud/services/rest/api/ProcessInstanceVariableAdminController.java
+++ b/activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-rest-api/src/main/java/org/activiti/cloud/services/rest/api/ProcessInstanceVariableAdminController.java
@@ -17,6 +17,9 @@ package org.activiti.cloud.services.rest.api;
 
 import org.activiti.api.process.model.payloads.RemoveProcessVariablesPayload;
 import org.activiti.api.process.model.payloads.SetProcessVariablesPayload;
+import org.activiti.cloud.api.model.shared.CloudVariableInstance;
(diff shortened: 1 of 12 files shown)


activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-rest-api/src/main/java/org/activiti/cloud/services/rest/api/ProcessInstanceVariableAdminController.java
```diff
@@ -25,11 +28,15 @@
 import org.springframework.web.bind.annotation.PutMapping;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
 
 @RequestMapping(value = "/admin/v1/process-instances/{processInstanceId}/variables",
         produces = {MediaTypes.HAL_JSON_VALUE, MediaType.APPLICATION_JSON_VALUE})
 public interface ProcessInstanceVariableAdminController {
 
+    @RequestMapping(method = RequestMethod.GET)
```
mteodori, line 37: It looks this fix is also adding new functionality, can we reference some issue or add an explanation in the PR? Thanks

balsarori: it just exposes an existing functionality that wasn't exposed before.

PR #651, opened 2022-03-23
AAE-7173 user and group search

jesty:
This pull request aims to expose user and group search rest services in order to replace the direct access to Keycloak Rest API from the UI. In this way, in the near future, we could provide implementations for other IAMs.

In order to do this, we are using Feign instead of using the Keycloak Rest Client because it has some vulnerabilities and we are going to remove it from our dependencies. 

The REST API is exposed in Modeling Service and in Runtime Bundle Service, we need to expose it twice because we need it in ADW to use it in the People Widget and in Modeling App and Admin App in order to assign users/groups to a task or an application.


jsokolowskii (approved):
LGTM

erdemedeiros:
Ref https://github.com/Activiti/Activiti/issues/3916

PR #656, opened 2022-03-28
[AAE-7855] variables in process definition

jsokolowskii:
This PR is adding query param `include`, which now only takes `variables` value. If this param is in the request, the endpoint returns displayable variable definitions for each process definition.

### DIFF
--- a/activiti-cloud-api/activiti-cloud-api-process-model/src/main/java/org/activiti/cloud/api/process/model/CloudProcessDefinition.java
+++ b/activiti-cloud-api/activiti-cloud-api-process-model/src/main/java/org/activiti/cloud/api/process/model/CloudProcessDefinition.java
@@ -16,8 +16,16 @@
 package org.activiti.cloud.api.process.model;
 
 import org.activiti.api.process.model.ProcessDefinition;
+import org.activiti.api.process.model.VariableDefinition;
 import org.activiti.cloud.api.model.shared.CloudRuntimeEntity;
 
+import java.util.List;
+
 public interface CloudProcessDefinition extends CloudRuntimeEntity, ProcessDefinition {
+
+    @Override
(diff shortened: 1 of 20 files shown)


activiti-cloud-api/activiti-cloud-api-process-model/src/main/java/org/activiti/cloud/api/process/model/CloudProcessDefinition.java
```diff
@@ -16,8 +16,16 @@
...
 
 import org.activiti.api.process.model.ProcessDefinition;
+import org.activiti.api.process.model.VariableDefinition;
 import org.activiti.cloud.api.model.shared.CloudRuntimeEntity;
 
+import java.util.List;
+
 public interface CloudProcessDefinition extends CloudRuntimeEntity, ProcessDefinition {
 
+    @Override
```
jesty: Why do we need it?

jsokolowskii: I've added `getVariableDefinitions()` to ProcessDefinition in activiti core, but in activiti-cloud ProcessInstanceEntity implements it (by implementing CloudProcessDefinition). This way I won't have to implement this method in entity, which would look weird.

jsokolowskii: The implementation moved to ExtendedCloudProcessDefinition.

jesty (approved):
LGTM

miguelruizdev (approved):

jesty (changes requested):
I would just check a doubt about the new dependency

jesty (approved):
done, LGTM

PR #661, opened 2022-04-04
[AAE-7467] swagger - remove custom schema name

jsokolowskii:
Removes custom naming of schemas for alternate rule types - automatic one works as expected.


jesty (approved):
LGTM

PR #696, opened 2022-05-16
AAE-7798 Remove Keycloak Adapter

miguelruizdev:
Removal of Keycloak dependencies from adapter in favor of JWT ones, suitable to implement different Auth providers, not restricted to Keycloak only.


activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-engine-util/pom.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jesty: Can we get rid of this module and put his utility in an existing library?

activiti-cloud-service-common/activiti-cloud-services-common-security-keycloak/src/main/java/org/activiti/cloud/services/common/security/config/CommonSecurityAutoConfiguration.java
```diff
@@ -0,0 +1,201 @@
...
+
+    @Override
+    protected void configure(HttpSecurity http) throws Exception {
+        authorizationConfigurer.configure(http);
+        http
+            .authorizeRequests().anyRequest().permitAll()
+            .and()
+            .cors()
+            .configurationSource(request -> {
+                CorsConfiguration corsConfiguration = new CorsConfiguration();
```
jesty: I forgot that we should review this configuration in order to be sure that we accept calls only from valid sources

activiti-cloud-service-common/activiti-cloud-services-test-security/pom.xml
```diff
@@ -20,8 +20,17 @@
...
+      <artifactId>spring-security-oauth2-client</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.security</groupId>
+      <artifactId>spring-security-oauth2-resource-server</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.jsonwebtoken</groupId>
+      <artifactId>jjwt</artifactId>
+      <version>0.9.1</version>
```
jesty: please move the version in the parent

activiti-cloud-service-common/activiti-cloud-services-test-security/pom.xml
```diff
@@ -31,5 +40,10 @@
       <groupId>org.junit.jupiter</groupId>
       <artifactId>junit-jupiter-api</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.springframework.security</groupId>
+      <artifactId>spring-security-oauth2-jose</artifactId>
+      <version>5.6.1</version>
```
jesty: please move the version in the parent

jesty (changes requested):

activiti-cloud-runtime-bundle-service/activiti-cloud-starter-runtime-bundle/src/main/java/org/activiti/cloud/starter/rb/configuration/RuntimeBundleSecurityAutoConfiguration.java
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.core.Ordered;
+import org.springframework.core.annotation.Order;
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.security.authentication.AbstractAuthenticationToken;
+import org.springframework.security.oauth2.client.registration.ClientRegistrationRepository;
+import org.springframework.security.oauth2.jwt.Jwt;
+
+@Configuration
+@Order(Ordered.HIGHEST_PRECEDENCE)
+public class RuntimeBundleSecurityAutoConfiguration extends CommonSecurityAutoConfiguration {
```
jesty: you don't need to extend it, otherwise, we are going to load the configuration twice. Please try to remove extends and annotate the method using bean, if it isn't loaded as expected add the @Primary annotation on it

jesty (changes requested):

activiti-cloud-service-common/activiti-cloud-services-common-security/src/main/java/org/activiti/cloud/identity/web/controller/IdentityManagementController.java
```diff
@@ -66,7 +66,7 @@ public List<Group> getGroups(@RequestParam(value = "search", required = false) S
     }
 
     @RequestMapping(value = "/roles", method = RequestMethod.GET)
-    public UserRoles getUserRoles(@AuthenticationPrincipal Principal principal) {
+    public UserRoles getUserRoles(@AuthenticationPrincipal Jwt principal) {
```
miguelruizdev, line 69: @jesty I had to change to a Jwt object to prevent the principal from being `null`.
As per the Javadocs, if the currently authenticated user does not match in format with the class marked as `@AuthenticationPrincipal`, the principal resolves to `null`.
See: https://docs.spring.io/spring-security/site/docs/4.0.3.RELEASE/apidocs/org/springframework/security/web/bind/support/AuthenticationPrincipalArgumentResolver.html
And:
https://stackoverflow.com/a/50622881/9658836

jesty (approved):
LGTM

PR #697, opened 2022-05-17
AAE-7813 handle empty properties and undefined process definition Id

jesty:
Process variable properties can be empty
Ref https://github.com/Activiti/Activiti/issues/3927


jsokolowskii (approved):
LGTM!

PR #699, opened 2022-05-18
#3934 fix NPE in RB schema

jsokolowskii:
This fixes a NPE which happens during api-docs generation for classes extending JsonDeserializer.
https://github.com/Activiti/Activiti/issues/3934


miguelruizdev (approved):
LGTM!

PR #706, opened 2022-05-25
AAE-8817 filter users by application keycloack

Giovanni007:
Keyclaok implementation to filter users by application
Ref: https://github.com/Activiti/Activiti/issues/3943

### DIFF
--- a/activiti-cloud-service-common/activiti-cloud-services-common-identity-keycloak/src/main/java/org/activiti/cloud/services/identity/keycloak/KeycloakManagementService.java
+++ b/activiti-cloud-service-common/activiti-cloud-services-common-identity-keycloak/src/main/java/org/activiti/cloud/services/identity/keycloak/KeycloakManagementService.java
@@ -15,7 +15,9 @@
  */
 package org.activiti.cloud.services.identity.keycloak;
 
+import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 import org.activiti.cloud.identity.GroupSearchParams;
@@ -27,11 +29,14 @@ import org.activiti.cloud.identity.model.User;
 import org.activiti.cloud.identity.model.UserRoles;
 import org.activiti.cloud.services.identity.keycloak.client.KeycloakClient;
 import org.activiti.cloud.services.identity.keycloak.mapper.KeycloakGroupToGroup;
+import org.activiti.cloud.services.identity.keycloak.mapper.KeycloakRoleMappingToRole;
 import org.activiti.cloud.services.identity.keycloak.mapper.KeycloakTokenToUserRoles;
 import org.activiti.cloud.services.identity.keycloak.mapper.KeycloakUserToUser;
+import org.activiti.cloud.services.identity.keycloak.model.KeycloakClientRepresentation;
 import org.activiti.cloud.services.identity.keycloak.model.KeycloakGroup;
 import org.springframework.security.oauth2.jwt.Jwt;
 import org.springframework.util.CollectionUtils;
+import org.apache.commons.lang3.StringUtils;
 
 public class KeycloakManagementService implements IdentityManagementService {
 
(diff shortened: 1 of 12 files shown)


activiti-cloud-service-common/activiti-cloud-services-common-identity-keycloak/src/main/java/org/activiti/cloud/services/identity/keycloak/KeycloakManagementService.java
```diff
@@ -42,28 +47,61 @@ public class KeycloakManagementService implements IdentityManagementService {
...
+            .collect(Collectors.toList());
+
+        if(!StringUtils.isEmpty(userSearchParams.getApplication())) {
+            return filterUsersInApplicationsScope(users, userSearchParams);
+        } else {
+            return filterUsers(users, userSearchParams);
+        }
+    }
+
+    private List<User> filterUsers(List<User> users ,UserSearchParams userSearchParams) {
```
miguelruizdev, line 79: just to double-check, the idea behind the `filterUsers` and `filterUsersInApplicationsScope` is that we provide both filtering options for single & multiple application?
in the same way we offer the possibility of setting `keycloak.use-resource-role-mappings` to both `false` and `true` ?

Giovanni007: If with "single application" you mean the `realm` level roles, yes.
In the basic scenario, we search for users and if we want to filter by their role we check the realm roles.
Instead, If we pass the parameter `application` then we return:

- the users that have access to the application (have at least one role in the Keycloack "client" role mapping)
- if we pass application and roles, then we return users that have that role in the client ignoring the REALM roles

miguelruizdev: Understood, thanks for the explanation 👌

miguelruizdev (approved):

activiti-cloud-service-common/activiti-cloud-services-common-identity-keycloak/src/main/java/org/activiti/cloud/services/identity/keycloak/KeycloakManagementService.java
mauriziovitale: Are we taking into account multiple apps ?

Giovanni007: No, the parameter `application` is a String, not a list.

balsarori (approved):
LGTM

activiti-cloud-service-common/activiti-cloud-services-common-identity-keycloak/src/main/java/org/activiti/cloud/services/identity/keycloak/KeycloakManagementService.java
```diff
@@ -42,28 +47,61 @@ public class KeycloakManagementService implements IdentityManagementService {
...
             .collect(Collectors.toList());
     }
 
+    private List<User> filterUsersInApplicationsScope(List<User> users ,UserSearchParams userSearchParams) {
+        String application = userSearchParams.getApplication();
+        String kClientId = getKeycloakClientId(application);
+        if(StringUtils.isEmpty(kClientId)) {
+            return Collections.emptyList();
+        }
+       users.forEach(user -> user.setApplicationRoles(
```
mauriziovitale, line 93: do we really need it?

mauriziovitale (changes requested):
check the comment

activiti-cloud-service-common/activiti-cloud-services-common-identity-keycloak/src/main/java/org/activiti/cloud/services/identity/keycloak/KeycloakManagementService.java
Giovanni007: I'm going to remove it in another PR

mauriziovitale (approved):

PR #725, opened 2022-06-13
fix(GH-3961): add missing Liquibase script in Query Service for M17 upgrade

igdianov:
After M17 version, the new behavior is now using composite key to assign activity instance id using process_instance_id, element_id and execution_id as the primary key. This PR ensure any upgraded from M16 (or older versions) to M17 (or newer versions) can keep running the ongoing processes without incurring in errors.

Part of https://github.com/Activiti/Activiti/issues/3961


erdemedeiros (approved):
LGTM!

andrea-ligios (approved):
LGTM!

PR #726, opened 2022-06-14
AAE-9131 Fix JWS Authentication error

Giovanni007:
Fix JWSAuthenticationManager returning null in case the Authentication is not a JWSAuthentication. In this way, the Spring ProviderManager will manage the wrong token exception received from the previous provider (JwtAuthenticationProvider).

Also fixing KeycloakResourceJwtAdapter to avoidNullPointerException when user have no resource roles


jesty (approved):
LGTM, please add a description to the PR

miguelruizdev (approved):

PR #728, opened 2022-06-15
Activiti/Activiti#3963 Fix implicit mappings conversion in modeling

### LINKED ISSUE #3963: Process validation error in modeling app when using mapping type
When validating a project that contains a process with a mapping using the mapping type, we see errors like this in the log:

```
com.fasterxml.jackson.databind.exc.InvalidFormatException: Cannot deserialize Map key of type `org.activiti.cloud.modeling.api.process.ServiceTaskActionType` from String "mappingType": not a valid representation, problem: (java.lang.IllegalArgumentException) No enum constant org.activiti.cloud.modeling.api.process.ServiceTaskActionType.MAPPINGTYPE
```

I think there is a problem here because the `mappingType` can’t be converted to the enumerated class ServiceTaskActionType The enumeration only contains the values (https://github.com/Activiti/activiti-cloud/blob/070c1d31aa41551f65d0269e2101b9a5212d8beb/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-api/src/main/java/org/activiti/cloud/modeling/api/process/ServiceTaskActionType.java#L27-L28 ):

- INPUT
- OUTPUT

This means that the mapping works correctly when we use the implicit mapping for a task because in that case, the mapping in the extensions looks like this:

```
"mappings": {
            "Task_0hkm1bw": {
                "inputs": {
                    "restUrl": {
                        "type": "variable",
                        "value": "str"
                    }
                }
            }
        }
```

But we can set the other types of implicit mappings like Map all inputs and outputs, map all inputs or map all outputs, that are placed in the `mappingType` attribute:

For example, for the option Map all inputs and outputs then the mapping looks like:

```
"mappings": {
            "Task_0hkm1bw": {
                "mappingType": "MAP_ALL"
            }
        }
```

For example, for the option Map all inputs variables then the mapping looks like:

```
"mappings": {
            "Task_0hkm1bw": {
                "mappingType": "MAP_ALL_INPUTS"
            }
        }
```

For example, for the option Map all outputs variables then the mapping looks like:

```
"mappings": {
            "Task_0hkm1bw": {
                "mappingType": "MAP_ALL_OUTPUTS"
            }
        }
```

As you can see, the mapping is not following the expected structure here https://github.com/Activiti/activiti-cloud/blob/070c1d31aa41551f65d0269e2101b9a5212d8beb/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-api/src/main/java/org/activiti/cloud/modeling/api/process/Extensions.java#L39  so it is not possible to convert the received JSON into a valid Extensions object

pmartinezga:
closes https://github.com/Activiti/Activiti/issues/3963

### DIFF
--- a/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-api/src/main/java/org/activiti/cloud/modeling/api/process/TaskMapping.java
+++ b/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-api/src/main/java/org/activiti/cloud/modeling/api/process/TaskMapping.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2017-2020 Alfresco Software, Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.activiti.cloud.modeling.api.process;
+
+import static com.fasterxml.jackson.annotation.JsonInclude.Include.NON_NULL;
+import java.util.HashMap;
+import java.util.Map;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+/**
+ * Email template mapping item
+ */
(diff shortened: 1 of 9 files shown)


activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-api/src/main/java/org/activiti/cloud/modeling/api/process/TaskMapping.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.Map;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+/**
+ * Email template mapping item
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(NON_NULL)
+public class TaskMapping {
```
miguelruizdev: maybe the name `TaskVariableMapping` is more specific?

miguelruizdev (approved):
LGTM!

PR #734, opened 2022-06-17
Add the possibility to replay a service task after failure

### LINKED ISSUE #3971: Add possibility to resend an integration request after a connector failure
After a connector failure, one should be able to send a new integration request, so that the process instance can continue its execution.

erdemedeiros:
A new REST endpoint is available to replay a service task based on the related execution id and flow-node id.
```
POST /rb/admin/v1/executions/{executionId}/replay/service-task
{
   "flowNodeId": "flowNodeId"
}
```

Fixes https://github.com/Activiti/Activiti/issues/3971


activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-rest-api/src/main/java/org/activiti/cloud/services/rest/api/RecoverController.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.springframework.hateoas.MediaTypes;
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+
+@RequestMapping(value = "/admin/v1/recover",
+    produces = {MediaTypes.HAL_JSON_VALUE, MediaType.APPLICATION_JSON_VALUE })
+public interface RecoverController {
```
igdianov: I think it would make sense to use better naming, i.e. `ServiceTaskAdminController` with base request endpoint `/admin/v1/service-task`. Then, use `POST /replay` path to map to replay method in the controller interface.

activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-rest-api/src/main/java/org/activiti/cloud/services/rest/api/ReplayServiceTaskRequest.java
```diff
@@ -0,0 +1,44 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.activiti.cloud.services.rest.api;
+
+import javax.validation.constraints.NotEmpty;
+
+public class ReplayServiceTaskRequest {
```
igdianov, line 21: I think this class should be moved to api module. We can also remove it in favor of using path variables to simplify , i.e. `POST /admin/v1/service-task/{executionId}/replay/{flowNodeId}`

activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-rest-api/src/main/java/org/activiti/cloud/services/rest/api/RecoverController.java
erdemedeiros: Yes, it makes sense. I'll refactor it.

erdemedeiros: After some discussion with the team, the new endpoint it's going to be:
```
POST /admin/v1/executions/{executionId}/replay/service-task
{
   "flowNodeId": "flowNodeId"
}
```
The parameter `executionId` is defining the resource, so it's going to be placed in the URL, while `flowNodeId` will go the the body, where we can eventually add extra information in the future (such as variables).

activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-rest-api/src/main/java/org/activiti/cloud/services/rest/api/ReplayServiceTaskRequest.java
erdemedeiros: As long we don't include a specific method in the Java API it should be fine to keep it at the same level as the controller. Am I missing something?

igdianov: I think it makes sense to keep it here.

igdianov (approved):

andrea-ligios (approved):
LGTM!

PR #745, opened 2022-06-29
AAE-9366 Fix firstname in Users API response to be camel case

Giovanni007:


Giovanni007:
Fixed with https://github.com/Activiti/activiti-cloud/pull/746/files

PR #753, opened 2022-07-11
Retrieve process variables by varName and processKey

jsokolowskii:
https://github.com/Activiti/Activiti/issues/3994

Process variables for the old processes (before 7.4.0) are not returned.
This happens because the system uses the variable definition id and before this field was not present in the database.

In order to retrieve all process variables we can combine variableName (already in the process_variable table) and processDefinitionKey (which should be migrated from process_instance to process_variable table).

The request will have the following form:

`http://{domain-name}/query/v1/process-instances?variableKeys={processDefinitionKey}%2F{variableName}`

e.g.

`http://localhost:8081/query/v1/process-instances?variableKeys=Process_90W_3nLpw%2FinitializedVar`

The response format will not change, but it will include both older and newly created process variables.


jesty (approved):
LGTM

PR #772, opened 2022-07-18
AAE-8993 avoid to create a new JWTAuthenticationToken and use the existing one

jesty:
The previous code was doing silly stuff, basically, it creates a new JWTAuthenticationToken from the existing missing username.


gicappa (approved):
LGTM!

PR #776, opened 2022-07-18
AAE-9961 create JWT  websocket token validator

jesty:
Abstract the existing Keycloak token validator and create a second implementation that could use for any JWT system.


gicappa (approved):
activiti-cloud-notifications-graphql-service/services/security/src/main/java/org/activiti/cloud/services/notifications/qraphql/ws/security/tokenverifier/jwt/JwtAccessTokenVerifier.java

PR #784, opened 2022-07-21
Use default request pagination values 

balsarori:
- Use default request pagination values 
- minor refactor to use Json serialisation instead of events classes `toString`


andrea-ligios (approved):
LGTM!

PR #806, opened 2022-08-12
AAE-10187 don't accept executable files in upload

jesty:
In this PR we are checking the magic number (or magic bytes) in order to don't accept executable files without relying only on file extension.

https://en.wikipedia.org/wiki/Magic_number_(programming)


activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/magicnumber/FileContentValidatorConfiguration.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+package org.activiti.cloud.services.modeling.validation.magicnumber;
+
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@EnableConfigurationProperties(FileMagicNumberList.class)
```
miguelruizdev, line 23: I like a lot the granularity of the new configuration files!
let's say goodbye to huge files difficult to understand

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/magicnumber/FileMagicNumber.java
```diff
@@ -0,0 +1,80 @@
...
+    private byte[] bytes;
+    private int offset;
+
+    public String getString() {
+        return string;
+    }
+
+    public void setString(String string) {
+        this.string = string;
+        if (string != null) {
```
miguelruizdev: could we enclose the logic of this setter into a private method giving it a descriptive name?
just to avoid the guess work when/if changing this logic in the future

miguelruizdev (changes requested):
just small comments, the rest looks so cool!

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/magicnumber/FileMagicNumber.java
jesty: good catch! done

miguelruizdev (approved):

PR #807, opened 2022-08-12
fix: optimize audit events handling by timestamp 

igdianov:
This PR add secondary sort by timestamp in EventHandlerOptimizer to ensure chronological handling order by event type to make sure that the semantical order is the right one, and in case of same value we order by timestamp


erdemedeiros (approved):
LGTM!

PR #819, opened 2022-08-26
fix: Update service task status in Query to STARTED on IntegrationRequestedEvent

igdianov:
The service task status will be updated to STARTED after IntegrationRequestedEvent is emitted and handled by Query Service.

Part of https://github.com/Activiti/Activiti/issues/4056


erdemedeiros (approved):
LGTM!

PR #820, opened 2022-08-30
AAE-10482 use client credentials when contacting keycloak

jesty:
KeycloakClient.java uses the propagated token to authenticate against Keycloak. It doesn't work when the invocation happens outside any web context, for instance when the authentication is required from an event consumer.

### DIFF
--- a/activiti-cloud-service-common/activiti-cloud-services-common-identity-keycloak/src/main/resources/keycloak-client.properties
+++ b/activiti-cloud-service-common/activiti-cloud-services-common-identity-keycloak/src/main/resources/keycloak-client.properties
@@ -1,6 +1,6 @@
 activiti.cloud.services.oauth2.iam-name=keycloak
 
-spring.security.oauth2.client.registration.keycloak.client-id=${activiti.keycloak.client-id}
+spring.security.oauth2.client.registration.keycloak.client-id=${activity.keycloak.client-id}
 spring.security.oauth2.client.registration.keycloak.client-secret=${activiti.keycloak.client-secret}
 spring.security.oauth2.client.registration.keycloak.authorization-grant-type=${activiti.keycloak.grant-type}
 spring.security.oauth2.client.registration.keycloak.scope=openid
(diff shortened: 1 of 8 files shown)


activiti-cloud-service-common/activiti-cloud-services-common-security/src/main/java/org/activiti/cloud/security/feign/ClientCredentialsAuthRequestInterceptor.java
```diff
@@ -0,0 +1,59 @@
...
+        this.clientRegistration = clientRegistration;
+    }
+
+    @Override
+    public Optional<String> getToken() {
+        Optional<String> result;
+        if (AuthorizationGrantType.CLIENT_CREDENTIALS.equals(clientRegistration.getAuthorizationGrantType())) {
+            OAuth2AuthorizeRequest oAuth2AuthorizeRequest = OAuth2AuthorizeRequest
+                .withClientRegistrationId(clientRegistration.getRegistrationId())
+                .principal("activiti")
```
miguelruizdev, line 47: how do we know 'activiti' is the right value?
is it always present?

activiti-cloud-service-common/activiti-cloud-services-common-identity-keycloak/src/main/resources/keycloak-client.properties
```diff
@@ -1,6 +1,6 @@
 activiti.cloud.services.oauth2.iam-name=keycloak
 
-spring.security.oauth2.client.registration.keycloak.client-id=${activiti.keycloak.client-id}
+spring.security.oauth2.client.registration.keycloak.client-id=${activity.keycloak.client-id}
```
scaccia: Why did you use `activity` with `y` at the end?

activiti-cloud-service-common/activiti-cloud-services-common-security/src/main/java/org/activiti/cloud/security/feign/ClientCredentialsAuthRequestInterceptor.java
jesty: is a placeholder, the principal is mandatory but basically not used in this context.

activiti-cloud-service-common/activiti-cloud-services-common-identity-keycloak/src/main/resources/keycloak-client.properties
jesty: it is a typo, thanks!

miguelruizdev (approved):

PR #821, opened 2022-08-31
[AAE-10480] : implementation find Identity Users By Group Id

scaccia:
- create new interface `IdentityService` with method `findUsersByGroupName`
- map new Feign Keycloak API: `/groups/{groupId}/members`
- implement `findUsersByGroupName` in `KeycloakManagementService`

### DIFF
--- a/activiti-cloud-service-common/activiti-cloud-services-common-identity-keycloak/src/test/java/org/activiti/cloud/services/identity/keycloak/KeycloakClientIT.java
+++ b/activiti-cloud-service-common/activiti-cloud-services-common-identity-keycloak/src/test/java/org/activiti/cloud/services/identity/keycloak/KeycloakClientIT.java
@@ -42,6 +42,8 @@ import org.testcontainers.shaded.org.awaitility.Awaitility;
 @ContextConfiguration(initializers = {KeycloakContainerApplicationInitializer.class})
 public class KeycloakClientIT {
 
+    public static final String HR_GROUP_ID = "60c6753b-4f1c-498a-96d2-be6790fae09c";
+    public static final String TEST_GROUP_NAME = "testgroup";
     @Autowired
     private KeycloakClient keycloakClient;
 
@@ -84,7 +86,7 @@ public class KeycloakClientIT {
 
     @Test
     public void should_searchGroups() {
-        List<KeycloakGroup> groups = keycloakClient.searchGroups("testgroup", 0, 50);
+        List<KeycloakGroup> groups = keycloakClient.searchGroups(TEST_GROUP_NAME, 0, 50);
 
         assertThat(groups).hasSize(1);
         assertThat(groups).extracting("name").contains("testgroup");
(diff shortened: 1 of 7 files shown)


activiti-cloud-service-common/activiti-cloud-services-common-identity-keycloak/src/test/java/org/activiti/cloud/services/identity/keycloak/KeycloakClientIT.java
```diff
@@ -313,4 +314,12 @@ public void should_getGroupsClientRoleMapping() {
         assertThat(groups).extracting("name").contains("salesgroup");
     }
 
+
+    @Test
+    public void should_getUsers_by_groupId() {
+        List<KeycloakUser> users = keycloakClient.getUsersByGroupId(HR_GROUP_ID);
+
+        assertThat(users).hasSizeGreaterThanOrEqualTo(1);
```
miguelruizdev, line 323: could we check at least the name of the user? (I assume it is `hruser` right?)

activiti-cloud-service-common/activiti-cloud-services-common-security/src/main/java/org/activiti/cloud/identity/IdentityRuntimeService.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.activiti.cloud.identity;
+
+import java.util.List;
+import org.activiti.cloud.identity.model.User;
+
+public interface IdentityRuntimeService {
```
jesty: Could you rename to IdentityService, please can leave a comment on IdentityManagerService saying that that interface should be used only in Identity Adapter Service and not for other purposes, while for genreral purpose we should use this class.

jesty (changes requested):

activiti-cloud-service-common/activiti-cloud-services-common-identity-keycloak/src/test/java/org/activiti/cloud/services/identity/keycloak/KeycloakClientIT.java
miguelruizdev: also, `assertThat(users).hasSize(1)` should be more accurate to what we want to check

scaccia: There's a couple of issues about that.

1. Based on configuration `activiti-realm.json` I know there are at least one user in HR_GROUP. In addition, on group HR_GROUP are performed some logic in other tests. So the only thing I can test is hasSizeGreaterThanOrEqualTo
image

2. Unfortunately, as showed in previous image, the other groups haven't got id

If necessary I can create a new group and new user only for testing this issue.

miguelruizdev: I see, now that I remember hruser and hradmin are inside that group, and maybe more
okay, it is safe to check it in the way you did
thanks for the clarification

miguelruizdev (approved):

jesty (approved):
LGTM

PR #824, opened 2022-09-05
Extend Identity Service interface 

miguelruizdev:
Extend Identity Service interface with more methods and implement it with Keycloak


jesty (approved):
LGTM

PR #825, opened 2022-09-06
AAE-10550 Return empty list for user search when group provided does not exist

miguelruizdev:
Return empty list for user search when group provided does not exist
Add exceptions to methods not used by the identity adapter layer


jesty (approved):
LGTM

PR #864, opened 2022-10-17
[AAE-10492] variables in admin process definitions 

### LINKED ISSUE #4112: List all available variables for admin [BE]
List all available variables in admin processDefinitions endpoint and allow to query variables at admin/v1/process-instances and admin/v1/tasks

LorenzoDiGiuseppe:
This PR is adding an `include` query param to `/admin/v1/process-definitions` endpoint, which only takes `variables` value at the moment. If it is set in a request, the endpoint returns all the displayable variable definitions for each process definition.
This PR is also adding the `variableKeys` query param to `/admin/v1/process-instances` and `/admin/v1/tasks` as an array of process parameters that allows to see the details of the specified parameters for all the process instances and tasks returned by the endpoints.


andrea-ligios (approved):
LGTM!

PR #879, opened 2022-10-25
fix: resolve Graphql schema build errors due to auto configuration race condition in Query Service

igdianov:
- [x] update graphql-jpa-query version to 0.5.3
- [x] replace duplicate ObjectScalar coercing with JavaScalars.GraphQLObjectCoercing implementation


erdemedeiros (approved):
LGTM!

PR #886, opened 2022-10-27
restrict process definition access to candidate starters when candidate starters is enabled

balsarori:
ref https://github.com/Activiti/Activiti/issues/4128


Giovanni007 (approved):
LGTM!

PR #904, opened 2022-11-17
AAE-11384: re-enable deployAtEnd (reverts 51676fbfa0022edba2c445241a7…

### LINKED ISSUE #4148: Fix maven deployAtEnd logic in Activiti modules
Changes done with https://github.com/Activiti/Activiti/issues/4032 led to rely on a single step to perform build and deployment.

But we’d expect deployment to be done “at end”, when all modules builds have successfully gone through.

This is not currently the case: the maven configuration needs to be checked to ensure the “deployAtEnd” option can be leveraged.

Jira ref: AAE-11384

atchertchian:
…5aefc942df5a0)

Fixes https://github.com/Activiti/Activiti/issues/4148

I did not find an explanation of why this was disabled on https://github.com/Activiti/activiti-cloud/pull/569 -> changing back to true so that deployment occurs when all the builds have gone through correctly

### DIFF
--- a/activiti-cloud-build/pom.xml
+++ b/activiti-cloud-build/pom.xml
@@ -480,7 +480,7 @@ limitations under the License.]]>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-deploy-plugin</artifactId>
         <configuration>
-          <deployAtEnd>false</deployAtEnd>
+          <deployAtEnd>true</deployAtEnd>
         </configuration>
       </plugin>
       <plugin>


erdemedeiros (approved):
@atchertchian , `depoyAtEnd` was initially disabled because it was not working properly in all the projects. Initially we thought that it was an issue with the plugin itself, but it turns out that the problem was related to a configuration in our project (you fixed that some time ago). So now it's safe to switch it on again.

PR #920, opened 2022-12-01
AAE-11572 use another conditional check, the previous does not work in RB

jesty:
I don't have to load this interceptor if I don't have any Oauth2 cliente configured. The previous check I was checking configuration properties and it doesn't work in RB.
Github issue: https://github.com/Activiti/Activiti/issues/4163


scaccia (approved):
LGTM

scaccia (approved):
LGTM

PR #923, opened 2022-12-02
[AAE-10563]: Fix Identity Service extension 

scaccia:


jesty (changes requested):
Please check if alfresco-process repo works with those changes.

scaccia:
@jesty The tests, using this new configuration, work correctly on  alfresco-process.
The only two classes that use the interface `IdentityService` are `EmailUserDetailsService` and `FindCandidateAssigneesFunction`. These classes will keep working like before.
Maybe we should refactor KeycloakManagementService because it implements methods used only internally to service

scaccia:
@jesty The tests, using this new configuration, work correctly on  alfresco-process.
The only two classes that use the interface `IdentityService` are `EmailUserDetailsService` and `FindCandidateAssigneesFunction`. These classes will keep working like before.
Maybe we should refactor KeycloakManagementService because it implements

jesty (approved):
LGTM

miguelruizdev (approved):

PR #940, opened 2022-12-23
Remove endpoints for creating and starting a created process instance

### LINKED ISSUE #4186: Remove endpoints for creating and starting a created Process Instance
We have to remove the unnecessary endpoints for creating a Process Instance and starting a previously created Process Instance since there is already an endpoint which does both actions and simplifies the interaction between front-end and back-end.

LorenzoDiGiuseppe:
This PR removes the endpoints for creating and starting a previously created Process Instance from RB. In fact, there is already the endpoint: `POST: /v1/process-instances` which creates and starts a Process Instance basing on the payload of the request.

This PR closes Activiti/Activiti#4186.


jsokolowskii (approved):
LGTM!

VitoAlbano (approved):

PR #944, opened 2022-12-30
[AAE-11547] - Replace Keycloak Client using Feign in Activiti

scaccia:
Ref: https://github.com/Activiti/Activiti/issues/4267

- Open point: it would be useful to split the keycloak feign client https://github.com/Activiti/activiti-cloud/blob/689e7abc4b32160d0aa8a5229bb224e0414098dd/activiti-cloud-service-common/activiti-cloud-services-common-identity-keycloak/src/main/java/org/activiti/cloud/services/identity/keycloak/client/KeycloakClient.java#L38.


jesty (approved):
LGTM

PR #959, opened 2023-01-10
fix vuln apache-mime4j

balsarori:
The vulnerable version of apache-mime4j-storage  is coming from resteasy. Latest 6.x resteasy version (currently we're using 4.7.5.Final) uses a newer of apache-mime4j-storage which isn't vulnerable. I tried to update to the latest resteasy version but faced compatibility issues. For now, I updated the version manually in our pom.xml.


gicappa (approved):
LGTM!

PR #969, opened 2023-01-13
fix: search users filtering by groups

Giovanni007:
User search filtered by a list of group is not working as expected. Current implementation returns all the user that are member of at least one of the groups passed as search parameter. The expected behaviour is to return only users that are member of  all the group passed in the search parameter.
ref. https://github.com/Activiti/Activiti/issues/4212


miguelruizdev (approved):
LGTM

PR #973, opened 2023-01-18
Disable Zipkin reporting

### LINKED ISSUE #4219: Disable Zipkin by default
It would be better disabling Zipkin by default to avoid error logs, since it is not always required by applications.

LorenzoDiGiuseppe:
This PR closes Activiti/Activiti#4219 disabling Zipkin.


jsokolowskii (approved):
LGTM

gicappa (approved):
LGTM!

PR #985, opened 2023-01-27
AAE-7800 fix keycloak client test reals configuration was wrong #984

jesty:
It was working by chance using the TokenRelayInterceptor that shouldn't be used.


scaccia (approved):
LGTM

PR #987, opened 2023-01-30
[AAE-11547] - add to getAllUsers method the maximum results size parameter

scaccia:
Ref: https://github.com/Activiti/Activiti/issues/4266


jesty (approved):
LGTM

PR #989, opened 2023-02-01
AAE-12447 extract imageutils to common utils

jsokolowskii:
https://github.com/Activiti/Activiti/issues/4232

We should have an ImageUtils class in the common utils which performs the conversion from SVG to PNG (and maybe more in the future).


LorenzoDiGiuseppe (approved):
LGTM

PR #995, opened 2023-02-07
Enable ConsumerConnector for lambdas and method references

LorenzoDiGiuseppe:
This PR allows using `ConsumerConnector` functional interface with lambdas and method references.
In fact, the interface introduced in Activiti/Activiti#4234 suffered of type erasure when used outside class implementations.


jsokolowskii (approved):
LGTM

PR #999, opened 2023-02-08
AAE-12448 convert svg to png for model extensions

jsokolowskii:
https://github.com/Activiti/Activiti/issues/4241
In order to mitigate the XSS in SVG uploaded file we can convert it to PNG. It applies to all extensions nodes.


LorenzoDiGiuseppe (approved):
LGTM

PR #1001, opened 2023-02-09
AAE-12486 sanitize files in model content

jsokolowskii:
https://github.com/Activiti/Activiti/issues/4241
Sanitizes FileContent containing SVG (potentially malicious) which is converted to PNG.


LorenzoDiGiuseppe (approved):
LGTM

PR #1006, opened 2023-02-15
[AAE-11712] Validate models name while saving

graric:


pmartinezga:
Take a look at the build, it looks like your changes are collapsing with other existing tests

pmartinezga (commented):
It is surprising for me that we had tests covering the case of using underscore and using capital letters I need to investigate this before merging

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/main/java/org/activiti/cloud/services/modeling/rest/controller/ModelingRestExceptionHandler.java
```diff
@@ -87,6 +88,13 @@ public void handleModelNameConflictException(ModelNameConflictException ex,
         response.sendError(CONFLICT.value(), ex.getMessage());
     }
 
+    @ExceptionHandler(ModelNameInvalidException.class)
+    public void handleModelNameInvalidException(ModelNameInvalidException ex,
+                                                 HttpServletResponse response) throws IOException {
+        logger.error(ex.getMessage(), ex);
+        response.sendError(CONFLICT.value(), ex.getMessage());
+    }
```
pmartinezga: You are doing the same as the previous exception handler so I think it is better to merge the three we have for sending the conflict like this
```java
@ExceptionHandler({
    ModelNameConflictException.class,
    ModelNameInvalidException.class,
    ModelScopeIntegrityException.class
})
```

pmartinezga (approved):

pmartinezga (changes requested):
asking request just because the mergify bot will merge this otherwise

pmartinezga (approved):

graric:
Ref: https://github.com/Activiti/Activiti/issues/4273

PR #1011, opened 2023-02-20
[Activiti/Activiti#4252] When process is finished remaining tasks that are assigned or created will be cancelled

### LINKED ISSUE #4252: Terminated user tasks showing as ASSIGNED in query service

wojciech-piotrowiak:
https://github.com/Activiti/Activiti/issues/4252

closes https://github.com/Activiti/Activiti/issues/4252

### DIFF
--- a/activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/query/events/handlers/ProcessCompletedEventHandler.java
+++ b/activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/query/events/handlers/ProcessCompletedEventHandler.java
@@ -26,6 +26,8 @@ import org.activiti.api.process.model.events.ProcessRuntimeEvent;
 import org.activiti.api.task.model.Task.TaskStatus;
 import org.activiti.cloud.api.model.shared.events.CloudRuntimeEvent;
 import org.activiti.cloud.api.process.model.events.CloudProcessCompletedEvent;
+import org.activiti.cloud.api.task.model.impl.CloudTaskImpl;
+import org.activiti.cloud.api.task.model.impl.events.CloudTaskCancelledEventImpl;
 import org.activiti.cloud.services.query.model.ProcessInstanceEntity;
 import org.activiti.cloud.services.query.model.QueryException;
 import org.activiti.cloud.services.query.model.TaskEntity;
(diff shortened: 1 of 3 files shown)


activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/query/events/handlers/ProcessCompletedEventHandler.java
```diff
@@ -15,44 +15,62 @@
...
+  }
+
+  private void markAllChildTasksAsCancelledWhenTheyAreAssignedOrCreated(ProcessInstanceEntity processInstanceEntity) {
+    Predicate<TaskEntity> cancellableTasks = task -> TaskStatus.ASSIGNED.equals(task.getStatus())
+        || TaskStatus.CREATED.equals(task.getStatus());
+
+    Stream.ofNullable(processInstanceEntity.getTasks())
+        .flatMap(Set::stream)
+        .filter(cancellableTasks)
+        .peek(task -> task.setStatus(TaskStatus.CANCELLED))
```
wojciech-piotrowiak: I'm not sure if this solution will be stable in long period. Task cancelling having dedicated event handler so maybe we could create a new task canceled event with those tasks?
It might help for situation when task cancel logic would change.

pmartinezga: This sounds good but, do you think you are able to call the handler for cancelled tasks from here?

wojciech-piotrowiak: See two options:
1. Use TaskCancelledEventHandler, just create event and let it handle it (dependency have to be set in spring config) 
2. Just publish message to 'queryConsumer' which will be routed and then handled by specific handler class

IMHO better option might be 1. as we don't have many readers for the same event

pmartinezga: I don't like creating an event for this, because it is a "fake" event, is not an event produced by the RB and can be confusing if we have a bridge or some other service listening to the topic where we will be placing the event.

Is it possible to call the TaskCancelledEventHandler directly? (I think this is what you meant by option 1 right?) If it is, then I would go with this approach.

wojciech-piotrowiak: ok, change applied, cannot double check at this moment (https://hxpmock-apa.envalfresco.com/ is down for deployments)

wojciech-piotrowiak: @pmartinezga @graric Tests were successful

graric (approved):

pmartinezga (approved):

PR #965, opened 2023-01-12
Validation for Script Error boundaries

LorenzoDiGiuseppe:
This PR fixes Activiti/Activiti#4184 introducing a validate method that allows ignoring warnings.

`ModelService` has been extended with:
- methods for getting validation errors
- an `AggregateErrorValidationStrategy` that allows executing validation from all the configured validators without interrupting the validation chain

`AggregateErrorValidationStrategy` is particularly meaningful to avoid that an exception for a warning could stop the execution of further validators that could reveal an error in the model.

PR #1017, opened 2023-02-23
[Activiti/Activiti#4225] Validations can return errors as result

wojciech-piotrowiak:


activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-api/src/main/java/org/activiti/cloud/modeling/api/ModelValidator.java
```diff
@@ -15,42 +15,72 @@
...
-     * @return handled content type
-     */
-    String getHandledContentType();
+  /**
+   * Validate the given model.
+   *
+   * @param modelContent      content of the model to validate
+   * @param validationContext the validation context
+   */
+  void validate(byte[] modelContent,
```
LorenzoDiGiuseppe: I think we could implement the new functionality directly in the `validate` method. If the exceptions are required we could handle them in `ProjectService` and `ModelService`. It would simplify the code reducing the duplication and it would also remove the need of having an `AggregateErrorValidationStrategy`. WDYT?

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/JsonSchemaModelValidator.java
```diff
@@ -41,111 +43,145 @@
...
+    String jsonContent = new String(modelContent);
+    try {
+      log.debug("Validating json model content: " + jsonContent);
+      processExtensionJson = extractJsonObject(modelContent);
+      validateJson(processExtensionJson);
+    } catch (JSONException jsonException) {
+      log.debug("Syntactic model JSON validation errors encountered",
+                jsonException);
+      String problem = "Could not parse:" + jsonContent;
+      return List.of(new ModelValidationError(problem, jsonException.getMessage()));
```
LorenzoDiGiuseppe: I think we should probably differentiate the classes used for returning different errors for allowing the caller to identify what error it is, for example creating a `JsonValidationError` and so on. WDYT?

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/extensions/ExtensionsJsonSchemaValidator.java
```diff
@@ -31,43 +32,52 @@
...
-            throw new SemanticModelValidationException("Semantic model validation errors encountered: "
-                    + validationExceptions.stream().map(ModelValidationError::getDescription).collect(Collectors.joining(",")),
-                                                       validationExceptions);
-        }
+  private void validateExtensionsInContext(byte[] bytes,
+                                           ValidationContext validationContext) {
+    List<ModelValidationError> validationExceptions = getValidationErrors(convertBytesToModel(bytes),
+                                                                          validationContext);
+    if (!validationExceptions.isEmpty()) {
+      throw new SemanticModelValidationException("Semantic model validation errors encountered: "
```
LorenzoDiGiuseppe: We should probably transform this exception to a collection of `SemanticModelError`, if we are going to create different error classes as said above.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/JsonSchemaModelValidator.java
```diff
@@ -41,111 +43,145 @@
...
+    } catch (ValidationException validationException) {
+      log.debug("Semantic model validation errors encountered: " + validationException.toJSON(),
+                validationException);
+      return getValidationErrors(validationException, processExtensionJson);
```
LorenzoDiGiuseppe: I think we can try to reduce the code duplication between `validate` and `validateAndReturnErrors`, if we decide to keep both of them.

LorenzoDiGiuseppe (commented):
I think we could avoid using exceptions in order to simplify the validators and to expose easier validation errors to the caller. In this way, we should also be able to remove the `ValidationStrategy` added in Activiti/activiti-cloud#965 as workaround for the same problem.
Following some comments.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/JsonSchemaModelValidator.java
wojciech-piotrowiak: Sounds good to create subclasses

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/extensions/ExtensionsJsonSchemaValidator.java
wojciech-piotrowiak: Good idea

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/JsonSchemaModelValidator.java
wojciech-piotrowiak: Seen this problem so moved core logic into methods that are reused in both cases. 
Not sure if there is a room for more changes, because those cases having different signatures, anyway I'm open for discussion

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-api/src/main/java/org/activiti/cloud/modeling/api/ModelValidator.java
wojciech-piotrowiak: Need to check logic in AggregateErrorValidationStrategy, but in general it sounds ok

wojciech-piotrowiak: After consideration, IMO refactor AggregateErrorValidationStrategy should be handled in other task to not increase scope of changes. At this point it should be doable to do other idea below: https://github.com/Activiti/activiti-cloud/pull/1017#discussion_r1116827591

wojciech-piotrowiak: AggregateErrorValidationStrategy removed

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/service/ModelServiceImpl.java
```diff
@@ -80,654 +77,657 @@
...
- * Business logic related to {@link Model} entities including process models, form models, connectors, data models and decision table models.
+ * Business logic related to {@link Model} entities including process models, form models, connectors, data models and
+ * decision table models.
  */
 @PreAuthorize("hasRole('ACTIVITI_MODELER')")
 @Transactional
 public class ModelServiceImpl implements ModelService {
 
-    private static final Logger logger = LoggerFactory.getLogger(ModelServiceImpl.class);
+  private static final Logger logger = LoggerFactory.getLogger(ModelServiceImpl.class);
```
LorenzoDiGiuseppe: It seems like you are using a different style from the project one (Google Java Style) and the new lines are 2 space indented instead of 4.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-api/src/main/java/org/activiti/cloud/modeling/api/SemanticModelError.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.activiti.cloud.modeling.api;
+
+public class SemanticModelError extends ModelValidationError {
```
LorenzoDiGiuseppe: `ModelValidationError` was used as model for semantic errors. We should probably define a new base class and extend from that one. Consider that `ModelValidationError` was already used in `BpmnCommonModelValidator`, I am afraid that this could change the scope of those validators, since they could return both `SemanticModelError` and `SyntacticModelError` now.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-api/src/main/java/org/activiti/cloud/modeling/api/SyntacticModelError.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.activiti.cloud.modeling.api;
+
+public class SyntacticModelError extends ModelValidationError {
```
LorenzoDiGiuseppe: See `SemanticModelError`.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/process/ProcessModelValidator.java
```diff
@@ -42,49 +38,62 @@
...
+    List<ModelValidationError> validationErrors = getModelValidationErrors(bytes,
+                                                                           validationContext);
 
-        if (!validationErrors.isEmpty()) {
-            String messageError = "Semantic process model validation errors encountered: " + validationErrors;
-            log.debug(messageError);
-            throw new SemanticModelValidationException(messageError,
-                                                       validationErrors);
-        }
+    if (!validationErrors.isEmpty()) {
```
LorenzoDiGiuseppe: See the comment on `ModelValidationError` as root for all the errors.
If we decide to keep it as root for error hierarchy, we should probably change the signature of `BpmnCommonModelValidator` since it could lead to unwanted implementations.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/JsonSchemaModelValidator.java
```diff
@@ -41,111 +45,145 @@
...
-                                                       getValidationErrors(validationException, processExtensionJson));
-        }
+  private final Logger log = LoggerFactory.getLogger(JsonSchemaModelValidator.class);
+
+  protected abstract SchemaLoader schemaLoader();
+
+  @Override
+  public void validate(byte[] modelContent,
+                       ValidationContext validationContext) {
+    internalValidation(modelContent, true);
```
LorenzoDiGiuseppe: We should probably remove `shouldThrowError` and avoid having a method which behavior changes basing on a boolean condition. We could instead decide inside `validate` if we have to throw an exception or not.
I would suggest to check if the collection contains a `SyntacticModelError` and throw a `SyntacticModelValidationException` otherwise if there are other errors to throw a `SemanticModelValidationException` and so on.
Since this logic seems common, we could probably also create an utility class for this kind of behavior to reuse it where it is required.

LorenzoDiGiuseppe (changes requested):
I left some comments below, in general I think that:
- the code style does not conform to the guidelines, please check your IDE configuration
- using `ModelValidationError` as root for all errors could lead to unwanted implementations
- there are some other minor improvements to do

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-api/src/main/java/org/activiti/cloud/modeling/api/ModelValidator.java
```diff
@@ -15,42 +15,72 @@
...
+
+  /**
+   * Validate the given model and its usage and return errors
+   *
+   * @param model             the model to validate
+   * @param modelContent      content of the model to validate
+   * @param validationContext the validation context
+   * @param validateUsage     validate the usage of a model
+   */
+  default Collection<ModelValidationError> validateAndReturnErrors(Model model,
```
erdemedeiros: Is this method used somewhere? It's a bit strange having a default implementation that's ignoring two of the provided parameters: it's forcing the caller to provide parameters that are completely ignored... in this case, the caller could simply call the method that have only two parameters.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-api/src/main/java/org/activiti/cloud/modeling/api/ModelValidator.java
```diff
@@ -15,42 +15,72 @@
...
+  void validate(byte[] modelContent,
+                ValidationContext validationContext);
+
+  /**
+   * Validate the given model and return errors
+   *
+   * @param modelContent      content of the model to validate
+   * @param validationContext the validation context
+   */
+  Collection<ModelValidationError> validateAndReturnErrors(byte[] modelContent,
```
erdemedeiros: I don't think adding a new method here is the way to go: I'd rather modify the existing `validate` method to return a list of errors instead of throwing an exception. Then it's up to the caller to decide what to do with them (throw an exception or do something else). If throwing an exception is needed in several places we could imagine to have a  simple wrapper that's calling the validation and throwing an exception in case of error. The callers can then use this wrapper when they need to throw an exception. Does it make sense?

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/test/java/org/activiti/cloud/services/modeling/config/GenericModelsConfiguration.java
```diff
@@ -36,219 +40,248 @@
...
-    };
+    @Override
+    public void validate(byte[] modelFile,
+                         ValidationContext validationContext) {
+      LOGGER.info("validate generic json content");
+    }
 
-    private ContentUpdateListener genericNonJsonContentUpdateListener = new ContentUpdateListener() {
+    @Override
+    public Collection<ModelValidationError> validateAndReturnErrors(byte[] modelContent,
```
erdemedeiros: Adding a new method, forces you to duplicate part of the logic across all the implementations of the interface.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/service/ModelServiceImpl.java
wojciech-piotrowiak: Weird, check intelij and it should be google
Screenshot 2023-02-28 at 12 38 25

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/JsonSchemaModelValidator.java
```diff
@@ -41,111 +45,145 @@
...
+  }
+
+  @Override
+  public Collection<ModelValidationError> validateAndReturnErrors(byte[] modelContent,
+                                                                  ValidationContext validationContext) {
+    return internalValidation(modelContent, false);
+  }
+
+  private Collection<ModelValidationError> internalValidation(byte[] modelContent, boolean shouldThrowError) {
```
erdemedeiros: If we avoid adding the extra method `validateAndReturnErrors`, this problem is going to disappear as well, but in general a boolean parameter such as `shouldThrowError` is a code smell: It's going to force us to add a  lot of `if/else` blocks while we could have split the logic into two distinct methods and call one or the other depending on the desired behaviour. For instance, here you could have a method `validateAndReturnErrors ` returning the list of errors and the method `validate` that calls `validateAndReturnErrors` and throws an exception if the list is not empty.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/JsonSchemaModelValidator.java
wojciech-piotrowiak: There are two scenarios where caller expect exception and second when expect collection that might contain validation errors, thats why there is a boolean. 
So how you would like to decide to throw or not? See only option where code will always collect and then throw or not

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/process/ProcessModelValidator.java
```diff
@@ -42,49 +38,62 @@
...
+      throw new SemanticModelValidationException(messageError,
+                                                 validationErrors);
     }
+  }
 
-    @Override
-    public ModelType getHandledModelType() {
-        return processModelType;
-    }
+  private List<ModelValidationError> getModelValidationErrors(byte[] bytes, ValidationContext validationContext) {
```
erdemedeiros: I think `validate` would be a better name for this method, as it's doing more than getting the list of errors, it's actually performing the validation which is returning the list of errors.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-api/src/main/java/org/activiti/cloud/modeling/api/ModelValidator.java
```diff
@@ -15,42 +15,72 @@
...
  * Business logic related with validation of a model
  */
 public interface ModelValidator extends ModelValidationErrorProducer {
 
-    /**
-     * Validate the given model.
-     * @param modelContent content of the model to validate
-     * @param validationContext the validation context
-     */
-    void validate(byte[] modelContent,
```
pmartinezga: Basically I think that what @LorenzoDiGiuseppe was asking here is that these `validate` methods (this one and the following) return a list of `ModelValidationError`

Then delegate in the implementations of the `ValidationStrategy` interface wether those errors are going to be thrown or not.

Doing this, we can avoid having to collect the exceptions like we do here:
https://github.com/Activiti/activiti-cloud/blob/b46c62242caa2accd9bd64e694d48621a29d6a15/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/service/utils/AggregateErrorValidationStrategy.java#L35-L44

Is this what your enhancement was about @LorenzoDiGiuseppe ??

LorenzoDiGiuseppe: Hi @pmartinezga, yes, this issue was for removing the needs to throw and to collect exceptions: https://github.com/Activiti/activiti-cloud/pull/965#discussion_r1088756734. 

In my opinion, we could probably remove the  `ValidationStrategy` since `ModelService` and `ProjectService` should be enough (there would be no more need to catch exceptions and aggregate them).

About the type of errors, I think just that we should not use `ModelValidationError` as super class since it is actually adopted for semantic errors only. If we decide to use it as base class for semantic and syntactic errors we should also change the implementations of `BpmnCommonModelValidator` because it would be like changing this interface signature.

I am including @erdemedeiros to the discussion, since he may have different ideas on this topic.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/service/ModelServiceImpl.java
```diff
@@ -80,654 +77,657 @@
...
+  }
+
+  private void validateModelContent(String modelType,
+                                    byte[] modelContent,
+                                    ValidationContext validationContext) {
+    modelContentService.findModelValidators(modelType)
+        .forEach(modelValidator -> modelValidator.validateModelContent(modelContent, validationContext));
+  }
+
+  private List<ModelValidationError> getModelContentValidationErrors(String modelType,
```
erdemedeiros: I think the methods `getModelContentValidationErrors` and `getModelExtensionsValidationErrors` (including the public ones) should disappear: only the methods `validateModelContent` and `validateModelExtensions` should be kept after have been refactored to return a list of errors instead of throwing an exception.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/JsonSchemaModelValidator.java
```diff
@@ -41,111 +45,145 @@
...
+      validateJson(processExtensionJson);
+    } catch (JSONException jsonException) {
+      log.debug("Syntactic model JSON validation errors encountered",
+                jsonException);
+      if (shouldThrowError) {
+        throw new SyntacticModelValidationException(jsonException);
+      } else {
+        String problem = "Could not parse:" + jsonContent;
+        return List.of(new SyntacticModelError(problem, jsonException.getMessage()));
+      }
```
erdemedeiros: Maybe in this specific case, where we are not able even to parse the json file, it's fine throwing an exception directly, without connecting it in a list, as something really bad happened: the provide file is not a json.

erdemedeiros (changes requested):
I think the refactoring is taking the wrong direction: we should not have two parallel methods (one throwing an exception and one returning a list of errors) everywhere. We should have a single one (`validate`) that's returning a list of errors, then it's up to the caller to decide what to do with them.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-api/src/main/java/org/activiti/cloud/modeling/api/ModelValidator.java
erdemedeiros: Is `SyntacticModelError` used only in the case where we are not able to parse the json? If that's the case maybe we don't need it as per my comment [here](https://github.com/Activiti/activiti-cloud/pull/1017#discussion_r1124265548). And then, no need for specialisation, we only have `ModelValidationError` as before. WDYT?

LorenzoDiGiuseppe: AFAIK, at the moment it is used just for mapping `JSONException` and when there is a problem converting `byte[]` to a model or model extension: [`ProcessExtensionsModelValidator`](https://github.com/Activiti/activiti-cloud/blob/9796c3a36c422613c42051df017bf7dc81e4330a/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/extensions/ProcessExtensionsModelValidator.java#L124-L131), [`ExtensionsJsonSchemaValidator `](https://github.com/Activiti/activiti-cloud/blob/70202e7a5d612c1e44dffd2ec56df0b3c27e8cec/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/extensions/ExtensionsJsonSchemaValidator.java#L65-L72). So it could still be good to keep throwing them and reduce the changes.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/test/java/org/activiti/cloud/services/modeling/config/GenericModelsConfiguration.java
wojciech-piotrowiak: On beginning assumed that it would be good to keep backward compatibility and just introduce a new method. If this is not a concern then I will just refactor existing method to return errors

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/JsonSchemaModelValidator.java
wojciech-piotrowiak: Ok, will throw this error

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/JsonSchemaModelValidator.java
wojciech-piotrowiak: with single method this param will not be used

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/test/java/org/activiti/cloud/services/modeling/config/GenericModelsConfiguration.java
erdemedeiros: You're right, it's always good to keep in mind backward compatibility, but in this specific case, those are not public APIs meant to be used by other people, so I think it's fine to refactor it as it better fits for us.

LorenzoDiGiuseppe (approved):
LGTM

pmartinezga (approved):

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/JsonSchemaModelValidator.java
```diff
@@ -66,50 +68,55 @@ public void validate(byte[] bytes,
             throw new SemanticModelValidationException(validationException.getMessage(),
                                                        getValidationErrors(validationException, processExtensionJson));
```
erdemedeiros: Shouldn't this one return the list of errors stead of throwing a SemanticModelValidationException?

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/extensions/ExtensionsJsonSchemaValidator.java
```diff
@@ -31,30 +33,33 @@
...
+                                                     ValidationContext validationContext) {
         super.validate(bytes,
                        validationContext);
 
         if (!validationContext.isEmpty()) {
             validateExtensionsInContext(bytes,
-                                         validationContext);
+                                        validationContext);
         }
+        return Collections.emptyList();
```
erdemedeiros: Shouldn't this be aggregating the list of errors from `super.validate` with `validateExtensionsInContext` and returning this instead of  throwing `SemanticModelValidationException` inside `validateExtensionsInContext` and returning empty list here?

erdemedeiros (changes requested):
@wojciech-piotrowiak thank you for applying the changes. It's looking much better now.
There are only two points left where I think we should be returning the list of errors instead of throwing an exception. Please check my comments.

erdemedeiros (approved):
LGTM!

atchertchian:
If this PR shows conflicts, here are instructions to help rebasing this work branch after reformat at https://github.com/Activiti/activiti-cloud/commit/814a9190ff60af86086e8d111660146569ef6e19:
```
git rebase c845f509770e38aa9663a97cb2f9bcd3e5d39e28
git rebase --strategy-option=theirs --exec 'pre-commit run --from-ref HEAD~ --to-ref HEAD' 814a9190ff60af86086e8d111660146569ef6e19
git rebase develop
```

The last rebase should be omitted if another reformat has happened (similar instructions should be followed for rebase over next reformat).

See more details at https://github.com/Activiti/Activiti/issues/4277.

PR #1020, opened 2023-02-24
AAE-12725 - add model name to modelContent validation

Giovanni007:
Required by https://github.com/Alfresco/alfresco-modeling-service/pull/611
Ref: https://github.com/Activiti/Activiti/issues/4261

### DIFF
--- a/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-api/src/main/java/org/activiti/cloud/modeling/api/ModelContentValidator.java
+++ b/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-api/src/main/java/org/activiti/cloud/modeling/api/ModelContentValidator.java
@@ -45,16 +45,4 @@ public interface ModelContentValidator extends ModelValidator {
                                       boolean validateUsage) {
         validate(model, modelContent, validationContext, validateUsage);
     }
-
-    /**
-     * Validate the given model content.
-     * @param modelName         the name of the model to validate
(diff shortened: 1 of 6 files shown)


pmartinezga:
I don't fully understand why we need a new method for this, is not enough with the one existing?
```java
validateModelContent(Model model,
                                      byte[] modelContent,
                                      ValidationContext validationContext,
                                      boolean validateUsage)
```
As in the `Model` parameter you already have the name

Giovanni007:
> I don't fully understand why we need a new method for this, is not enough with the one existing?
> 
> ```java
> validateModelContent(Model model,
>                                       byte[] modelContent,
>                                       ValidationContext validationContext,
>                                       boolean validateUsage)
> ```
> 
> As in the `Model` parameter you already have the name

@pmartinezga My concern is that the validation is done [here](https://github.com/Activiti/activiti-cloud/pull/1020/files#diff-c7334a70aba77a84775f7681f6f8847526bccaee25dc2de124e4f7ff41190aa2L632) using `modelValidator.validateModelContent(modelContent, validationContext));
`

I supposed that changing it with `validateModelContent(Model, ...)`  is a breaking change because a modelValidator could have a different implementation for the two methods. So it means that, if I use it, I have to check every single **modelValidator** implementations that currently we have in our code to compare the two methods, than change or implement  `validateModelContent(Model, ...)` for each `modelValidator`. But that also doing so I  still risk to break existing customization made by customers. 
Do you think this is a  legitimate concern? 

My solution to this concern was to implement the new method, so that I can "relay" on the existing implementation as default, and override it only for the modelValidator I need.

pmartinezga (approved):

PR #1028, opened 2023-03-02
4269 - Improve cache of OAuth2 UserInfo Endpoint using sid

scaccia:
Ref: https://github.com/Activiti/Activiti/issues/4269


miguelruizdev (approved):
LGTM

jesty (commented):
we don't have the Sid/sub problem here?

scaccia:
> we don't have the Sid/sub problem here?

AFAIK It arrives a token exchange token that always has the sid.

PR #1035, opened 2023-03-06
AAE-12923 add an authenticate method that accept a client secret

jesty:


activiti-cloud-acceptance-tests/activiti-cloud-acceptance-tests-shared/src/main/java/org/activiti/cloud/acc/shared/service/AuthenticationService.java
```diff
@@ -31,4 +31,12 @@ AuthToken authenticate(@Param("client_id") String clientId,
                            @Param("grant_type") String grantType,
                            @Param("username") String username,
                            @Param("password") String password);
+
+    @RequestLine("POST")
+    @Headers("Content-Type: application/x-www-form-urlencoded")
+    AuthToken authenticate(@Param("client_id") String clientId,
+                           @Param("client_secret") String clientSecret,
```
miguelruizdev, line 38: can't we add the `client-secret` param to the existing method?

scaccia (approved):
LGTM

activiti-cloud-acceptance-tests/activiti-cloud-acceptance-tests-shared/src/main/java/org/activiti/cloud/acc/shared/service/AuthenticationService.java
jesty: I prefer to have a new one in order to avoid to set it to empty or null where it isn't required

miguelruizdev (approved):
LGTM

PR #1066, opened 2023-04-13
AAE-13346 remove fix spring web version

jsokolowskii:
With the migration to spring-boot 2.7.10 we can remove the fix version of spring-web as a higher version is already included.


LorenzoDiGiuseppe (approved):
LGTM

PR #1072, opened 2023-04-19
Fix execute command PropagateCloudBpmnErrorCmd

### LINKED ISSUE #4305: Fix the status update of a service task when an error occurs
When an error occurs, the event update is not sent during a service task and the status is not changed to ERROR.

mavotto:
This PR fixes https://github.com/Activiti/Activiti/issues/4305.

When there is a `CloudBpmnError` the command `new PropagateCloudBpmnErrorCmd(integrationError, execution)` is added, which in this case throws an exception which is caught, then when we re-execute `managementService.executeCommand(CompositeCommand.of(commands.toArray(Command[]::new)));` at the end of the method we still have `PropagateCloudBpmnErrorCmd` in the command list, and then the exception is thrown again, which in this case is not caught, so the event to update the status of the service activity is not sent.

To solve this bug, the command list is cleaned in the catch 
Added unit test for this scenario

### DIFF
--- a/activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-connectors/src/main/java/org/activiti/services/connectors/channel/ServiceTaskIntegrationErrorEventHandler.java
+++ b/activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-connectors/src/main/java/org/activiti/services/connectors/channel/ServiceTaskIntegrationErrorEventHandler.java
@@ -18,6 +18,7 @@ package org.activiti.services.connectors.channel;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 import org.activiti.api.process.model.IntegrationContext;
 import org.activiti.cloud.api.process.model.CloudBpmnError;
 import org.activiti.cloud.api.process.model.IntegrationError;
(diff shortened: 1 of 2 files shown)


miguelruizdev (approved):
LGTM

activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-connectors/src/main/java/org/activiti/services/connectors/channel/ServiceTaskIntegrationErrorEventHandler.java
```diff
@@ -117,6 +118,12 @@ public void receive(IntegrationError integrationError) {
                             return;
                         } catch (Throwable cause) {
                             LOGGER.error("Error propagating CloudBpmnError: {}", cause.getMessage());
+                            // cleaned the commands list from PropagateCloudBpmnErrorCmd and AggregateIntegrationErrorReceivedClosingEventCmd
```
jesty, line 121: instead of a comment, can we extract a method with a proper name? thanks

jesty (changes requested):

activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-connectors/src/main/java/org/activiti/services/connectors/channel/ServiceTaskIntegrationErrorEventHandler.java
```diff
@@ -150,4 +153,11 @@ public void receive(IntegrationError integrationError) {
             managementService.executeCommand(CompositeCommand.of(commands.toArray(Command[]::new)));
         }
     }
+
+    private static List<Command<?>> restoreCommandList(List<Command<?>> commands) {
```
jesty: can you remove static

jesty (changes requested):

activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-connectors/src/main/java/org/activiti/services/connectors/channel/ServiceTaskIntegrationErrorEventHandler.java
mavotto: Done

jesty (approved):
LGTM

PR #1075, opened 2023-04-25
[AAE-14012] new content type

wojciech-piotrowiak:


activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/resources/schema/process-extensions-schema.json
```diff
@@ -102,6 +102,7 @@
             "json",
             "file",
             "folder",
+            "content",
```
pmartinezga, line 123: You need to add also an `if/then` statement to the `allOf` object in line `140` that allows the content to be an object or a valid JUEL expression

```json
        {
          "if": {
            "properties": {
              "type": {
                "const": "content"
              }
            }
          },
          "then": {
            "properties": {
              "value": {
                "anyOf": [
                  {
                    "type": "object",
                    "message": {
                      "type": "Mismatch value type - {{name}}({{id}}). Expected type is content"
                    }
                  },
                  {
                    "type": "string",
                    "pattern": "^\\$\\{(.*)[\\}]$",
                    "message": {
                      "pattern": "Value format in {{name}}({{id}}) is not a valid expression",
                      "type": "Value format in {{name}}({{id}}) is not a valid expression"
                    }
                  }
                ]
              }
            }
          }
        }
```

pmartinezga (changes requested):

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/resources/schema/process-extensions-schema.json
wojciech-piotrowiak: done, thanks

pmartinezga (approved):

PR #1087, opened 2023-05-19
AAE-14299 extend keycloak client api for pagination

jsokolowskii:
https://github.com/Activiti/Activiti/issues/4327
The keycloak API endpoints `clients/{id}/roles/{role-name}/users` and `/groups/{groupId}/members` handle pagination and it should be reflected in KeycloakClient. It enables projects based on activiti-cloud to handle pagination.


LorenzoDiGiuseppe (approved):
LGTM

PR #1096, opened 2023-05-25
Multi-thread Maven Build

### LINKED ISSUE #4349: Test should be executed in parallel to reduce build time
It should be possible to execute unit tests in parallel and we could introduce multi-threaded builds to further improve the build time of `activiti-cloud`.

LorenzoDiGiuseppe:
This PR fixes Activiti/Activiti#4349 as it configures the project for multithread builds:
- it introduces the profile `acceptance-tests` with activation on property `skipAcceptanceTests=false` which is the default value to exclude `serenity-maven-plugin` from the main aggregator since it was causing random failures in the following goals: `default-testCompile`, `test`, `integration-test`
- the Unit Tests are executed in parallel by classes, meaning that UT methods are executed sequentially inside a class avoiding conflicts
- nothing is changed for Integration Tests since they cannot be executed in parallel unless defining `@ResourceLocks` or `@Isolated` on test classes to avoid race-conditions (it would not improve the performances in the end, see: Activiti/activiti-cloud#1073)
- Integration Tests belonging to different modules are executed in parallel, so it has been required removing static service ports where defined to avoid errors
- `license-maven-plugin` has been upgraded to version 2.1.0 that includes the changes from the 2.0.1-alfresco.2 and it marks the goal `add-third-party` as `threadsafe`.

__NOTE__: the plugin `assertj-assertions-generator-maven-plugin` is not marked as `threadsafe`, but the [Mojo execution](https://github.com/assertj/assertj-assertions-generator-maven-plugin/blob/ae656e8caf99c7c27f725be8fa6b68b334b27ba1/src/main/java/org/assertj/maven/AssertJAssertionsGeneratorMojo.java#L209-L213) seems to be safe enough to use.


andrea-ligios (approved):
LGTM!

.github/workflows/main.yml
```diff
@@ -75,7 +75,7 @@ jobs:
 
       - name: Maven Build and Test
         shell: bash
-        run: mvn -DnoAcceptanceTests=true install ${{ env.MAVEN_CLI_OPTS}}
+        run: mvn -DnoAcceptanceTests=true -DunitTests.parallel=true -T 1C install ${{ env.MAVEN_CLI_OPTS}}
```
erdemedeiros: I know that this was not introduced by this PR,  but what do you think about renaming `noAcceptanceTests` to `skipAcceptanceTests`?

activiti-cloud-runtime-bundle-service/activiti-cloud-starter-runtime-bundle-it/src/test/resources/application.properties
```diff
@@ -7,7 +7,7 @@ spring.jmx.enabled=false
 spring.rabbitmq.host=localhost
 
 # Shared H2 database instance
-spring.datasource.url=jdbc:h2:tcp://localhost:9090/mem:mydb
+# spring.datasource.url=jdbc:h2:tcp://localhost:9090/mem:mydb
```
erdemedeiros: Please, remove commented code.

erdemedeiros (changes requested):
Great job, @LorenzoDiGiuseppe !
There is one small change needed and one optional.

activiti-cloud-runtime-bundle-service/activiti-cloud-starter-runtime-bundle-it/src/test/resources/application.properties
LorenzoDiGiuseppe: Done, thank you.

.github/workflows/main.yml
LorenzoDiGiuseppe: Done, thank you.

erdemedeiros (approved):
LGTM!

PR #1108, opened 2023-06-02
AAE-10754 Migrate Activiti maven projects to JDK 17

igdianov:
- [x] Update build scripts to use JDK 17
- [x] Update Java version to 17
- [x] AAE-10797 Migrate Activiti docker images to JDK 17
- [x] Deprectated preview branch support from CI build replaced by `preview` label

Depends on
- https://github.com/Activiti/Activiti/pull/4337

### DIFF
--- a/activiti-cloud-acceptance-tests/pom.xml
+++ b/activiti-cloud-acceptance-tests/pom.xml
@@ -17,7 +17,7 @@
     <module>activiti-cloud-acceptance-tests-core</module>
   </modules>
   <properties>
-    <serenity.version>3.6.12</serenity.version>
+    <serenity.version>3.6.23</serenity.version>
     <serenity-jbehave.version>1.46.0</serenity-jbehave.version>
     <feign-form.version>3.8.0</feign-form.version>
     <batik.version>1.16</batik.version>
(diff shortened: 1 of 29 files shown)


activiti-cloud-acceptance-tests/pom.xml
```diff
@@ -17,11 +17,12 @@
     <module>activiti-cloud-acceptance-tests-core</module>
   </modules>
   <properties>
-    <serenity.version>2.3.31</serenity.version>
+    <serenity.version>3.6.23</serenity.version>
```
atchertchian, line 20: I don't get why this version is specified multiple times: ideally versions should be shared on the root pom once and for all (and packages declaration in the dependencyManagement section).
I guess this cleanup should be done independently given the current status, I'm just posting a comment in case I missed something (?)
cc @erdemedeiros

activiti-cloud-examples/activiti-cloud-identity-adapter/Dockerfile
```diff
@@ -1,5 +1,7 @@
-#FROM adoptopenjdk/openjdk11:jdk-11.0.2.7-alpine-slim
-FROM openjdk:11.0.1-jre-slim-stretch
+FROM docker.io/library/alpine:3.17@sha256:124c7d2707904eea7431fffe91522a01e5a861a624ee31d03372cc1d138a3126
```
atchertchian: We're using `alpine:3.18@sha256:02bb6f428431fbc2809c5d1b41eab5a68350194fb508869a33cb1af4444c9b11` in other modules: should we align on the same version here too?

andrea-ligios (approved):
LGTM! Let's change the alpine version though

PR #1125, opened 2023-06-20
AAE-15318 ryuk disabled

jsokolowskii:
https://github.com/Activiti/Activiti/issues/4355
Disable ryuk for testcontainers reuse scenario (it is not needed as we delete the containers after the build is finished).


LorenzoDiGiuseppe (approved):
LGTM

PR #1136, opened 2023-06-28
[AAE-15491] Fix swagger generation

### LINKED ISSUE #4362: Fix swagger generation after updating Springdoc and Swagger
After updating Spring boot 3 and the required swagger libraries, all controllers with multiple methods for the same endpoint are generated with duplicate attributes, creating semantically incorrect swaggers that do not allow SDK generation.

mavotto:
This PR fixes https://github.com/Activiti/Activiti/issues/4362

- refactored all methods with the same name to avoid semantic errors in the swagger definition
- refactored variables that were creating an error in code generation
- added `hidden = true` for those controllers that have multiple methods for the same endpoint, this change resolves parameter duplication 

https://alfresco.atlassian.net/browse/AAE-15491


miguelruizdev (approved):

jesty (approved):
LGTM

PR #1104, opened 2023-05-31
[AAE-9137] Fix imported project version

RahulKumar1606:
Process version is showing 0.0.2 when importing a project. 
No matter what was the process version before, it will be 0.0.2 after import.

https://github.com/Activiti/Activiti/issues/4348

PR #1139, opened 2023-06-29
Revert "[AAE-9137] Fix imported project version (#1104)"

LorenzoDiGiuseppe:
This reverts commit b70a93a570cfa39e8cfbc3c3eda941fe61d343a8 for Activiti/Activiti#4348
The method is causing strange behavior by `spring-data` on tables `model`, `model_versions` and `model_version`, i.e. it leaves pending records in `model_version` that are not attached to any `model` causing foreign key violations in every delete of imported models/projects.


matteo4diani (approved):
LGTM 👍🏼

PR #1145, opened 2023-07-05
AAE-7344 changes added again because of the revert done previously

IshwarGupta1:
This is the PR having same changes as it was previously provided in PR-1117 of activiti-cloud for story AAE-7344
The changes in the activiti cloud affected some Integration tests of Modeling service and Igor Dianov reverted these changes previously because it was blocking release pipeline.
I have reviewed the issues and found out that the integration tests and those were affecting because according to my changes
if 
assignee -$initiator
candidate user & candidate group - both null

then validation should not happen and 400 should be thrown
I have fixed those changes but need to merge this PR to get the new jar version and re run tests for full confirmation.

the issues
Error:    ProcessValidationControllerIT.should_returnStatusNoContent_when_validatingAValidateProcessTextContentType:84 Status expected:<204> but was:<400>
Error:    ProcessValidationControllerIT.should_returnStatusNoContent_when_validatingAValidateSimpleProcess:68 Status expected:<204> but was:<400>
Error:    ReleaseControllerIT.should_returnStatusCreatedAndGivenNameAndComment_when_creatingAProjectReleaseWithPayload:1326 Range for response status value 400 expected:<SUCCESSFUL> but was:<CLIENT_ERROR>

### DIFF
--- a/activiti-cloud-modeling-service/activiti-cloud-acceptance-tests-modeling/src/main/java/org/activiti/cloud/acc/modeling/steps/ModelingModelsSteps.java
+++ b/activiti-cloud-modeling-service/activiti-cloud-acceptance-tests-modeling/src/main/java/org/activiti/cloud/acc/modeling/steps/ModelingModelsSteps.java
@@ -207,7 +207,9 @@ public class ModelingModelsSteps extends ModelingContextSteps<Model> {
 
     @Step
     public void checkCurrentModelValidation() throws IOException {
-        assertThat(validateCurrentModel()).extracting(Response::status).containsOnly(HttpStatus.SC_NO_CONTENT);
+        assertThat(validateCurrentModel())
+            .extracting(Response::status)
+            .containsOnly(HttpStatus.SC_NO_CONTENT, SC_BAD_REQUEST);
     }
 
     @Step
--- a/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/test/java/org/activiti/cloud/services/modeling/rest/controller/ModelControllerIT.java
+++ b/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/test/java/org/activiti/cloud/services/modeling/rest/controller/ModelControllerIT.java
@@ -523,7 +523,7 @@ public class ModelControllerIT {
 
         final ResultActions resultActions = mockMvc
             .perform(multipart("/v1/models/{model_id}/validate", processModel.getId()).file(file))
-            .andExpect(status().isNoContent());
+            .andExpect(status().isBadRequest());
     }
 
     @Test
--- a/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/test/java/org/activiti/cloud/services/modeling/rest/controller/ModelValidationControllerIT.java
+++ b/activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/test/java/org/activiti/cloud/services/modeling/rest/controller/ModelValidationControllerIT.java
@@ -114,7 +114,7 @@ public class ModelValidationControllerIT {
 
         mockMvc
             .perform(multipart("/v1/models/{model_id}/validate", processModel.getId()).file(file))
-            .andExpect(status().isNoContent());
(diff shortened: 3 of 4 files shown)


activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-service/src/main/java/org/activiti/cloud/services/modeling/validation/process/BpmnModelUserTaskAssigneeValidator.java
jsokolowskii, line 1: If I understand correctly, an exception will be thrown when the assignee is `$initiator`. As far as I know this is actually an intended behavior, as it will assign the task to the person initiating it. I feel this is not solving the bug in the referenced ticket, which is about problem with releasing a task in the runtime app. It should be probably fixed on runtime-bundle, not modeling service.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/test/java/org/activiti/cloud/services/modeling/rest/controller/ModelValidationControllerIT.java
```diff
@@ -114,7 +114,7 @@ public void should_returnStatusNoContent_when_validatingProcessModelWithValidCon
 
         mockMvc
             .perform(multipart("/v1/models/{model_id}/validate", processModel.getId()).file(file))
-            .andExpect(status().isNoContent());
+            .andExpect(status().isBadRequest());
```
jsokolowskii, line 117: The name of the test says `should_returnStatusNoContent_when_validatingProcessModelWithValidContent` while here the expected status is changed to `Bad Request`. The test case is actually testing the positive path and we should rather in case of an implementation change we should rather:

- fix the test case so that it still tests the positive path
- create a new test case for the `Bad Request` path

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-rest/src/test/java/org/activiti/cloud/services/modeling/rest/controller/ModelControllerIT.java
```diff
@@ -523,7 +523,7 @@ public void should_returnStatusNoContent_when_validatingProcessModelWithValidCon
 
         final ResultActions resultActions = mockMvc
             .perform(multipart("/v1/models/{model_id}/validate", processModel.getId()).file(file))
-            .andExpect(status().isNoContent());
+            .andExpect(status().isBadRequest());
```
jsokolowskii, line 526: This is the same situation as with the other test case with changed expectation. Please see the comment there.

activiti-cloud-modeling-service/activiti-cloud-acceptance-tests-modeling/src/main/java/org/activiti/cloud/acc/modeling/steps/ModelingModelsSteps.java
```diff
@@ -207,7 +207,9 @@ private FormData getFormData(Model model) throws IOException {
 
     @Step
     public void checkCurrentModelValidation() throws IOException {
-        assertThat(validateCurrentModel()).extracting(Response::status).containsOnly(HttpStatus.SC_NO_CONTENT);
+        assertThat(validateCurrentModel())
+            .extracting(Response::status)
+            .containsOnly(HttpStatus.SC_NO_CONTENT, SC_BAD_REQUEST);
```
jsokolowskii, line 212: This is a step used in acceptance tests to check a valid model (it is used on the step with `the model is valid` comment). We should not extend the expected response codes, since it can lead to false positive acceptance tests.

jsokolowskii (changes requested):
As I explained in the code comment, unfortunately I don't feel that this PR solves the issue with releasing a task at runtime. For a better understanding of the issue, please refer to the task creator. Also please make sure to fix tests and create new test cases for the new solutions.
Regarding the need to have a new version to test on a different project, there is a `preview` feature, that you can use in the future. You need to create a branch with `-preview` suffix and the PR build will create a SNAPSHOT version with maven and docker artifacts that can be used in other projects.

PR #1178, opened 2023-08-11
[AAE-15727] add and modify columns for name in project table

indrayanganguly:


jsokolowskii (approved):
Edit - changed review on @LorenzoDiGiuseppe comment, I think it has a good point.

activiti-cloud-modeling-service/activiti-cloud-services-modeling/activiti-cloud-services-modeling-api-impl/src/main/java/org/activiti/cloud/modeling/api/impl/ProjectImpl.java
```diff
@@ -62,12 +65,35 @@ public void setId(String id) {
...
     }
 
     @Override
     public void setName(String name) {
-        this.name = name;
+        setTechnicalName(name);
+
+        //TODO: Modify temporary code that sets display name same as technical name
+        setDisplayName(name);
+    }
```
LorenzoDiGiuseppe, line 77: I will leave just a comment since I will be off after today and I don't want to block the PR.
I think we should not set both names here, but just delegate to `setTechnicalName`. If it is required to set also the display name, I think it would be better doing it in another place to avoid having issues when get/set name will be deleted.
In addition, it would be easier implementing `getName` and `setName` with a `default` implementation in the interfaces to avoid repeating the same behavior.

activiti-cloud-modeling-service/activiti-cloud-acceptance-tests-modeling/src/main/java/org/activiti/cloud/acc/modeling/modeling/ModelingNamingIdentifier.java
```diff
@@ -50,7 +50,7 @@ public static ModelingNamingIdentifier<Project> projectNamed(String name) {
         return new ModelingNamingIdentifier<Project>(name) {
             @Override
             protected String getName(Project project) {
-                return project.getName();
+                return project.getTechnicalName();
```
VitoAlbano, line 53: The step to renaming the getTechnicalName that is not the final name yet should be done in the end Why it's already here?

jsokolowskii (changes requested):
Please check @LorenzoDiGiuseppe comment https://github.com/Activiti/activiti-cloud/pull/1178#discussion_r1291318262

VitoAlbano (changes requested):
This PR might actually be a super risk of breaking a lot of stuff , AS we pointed out in the ADR we discussed the first step is to add only the getDisplayName, not touching the getName.

PR #1185, opened 2023-08-23
AAE-15817 Refactor activiti-cloud-services-tracing transitive dependencies from starter modules

igdianov:
So, that it will be used directly in the example applications to avoid pulling transitive `micrometer-tracing` and `micrometer-tracing-bridge-brave` dependencies onto the class path when using application starter modules in AAE projects.


erdemedeiros (approved):
LGTM!

eromano (approved):

PR #1187, opened 2023-08-25
[AAE-15956] Write zip content to file

pmartinezga:
Enable the zip builder util to write the zip content to a file using a stream so we don't need to load the whole byte array in memory to do so.


scaccia (approved):
LGTM

PR #1245, opened 2023-11-13
AAE-18015 / AAE-18003 Duplicate key value violates unique constraints for process_instance_pkey / bpmn_activity_pkey

igdianov:
- [x] AAE-18015 Duplicate key value violates unique constraint "process_instance_pkey"
- [x] AAE-18003 Duplicate key value violates unique constraint "bpmn_activity_pkey"

### DIFF
--- a/activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/query/events/handlers/BPMNSequenceFlowTakenEventHandler.java
+++ b/activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/query/events/handlers/BPMNSequenceFlowTakenEventHandler.java
@@ -31,15 +31,46 @@ public class BPMNSequenceFlowTakenEventHandler implements QueryEventHandler {
     private static final Logger logger = LoggerFactory.getLogger(BPMNSequenceFlowTakenEventHandler.class);
 
     private final EntityManager entityManager;
+    private final EntityManagerFinder entityManagerFinder;
 
-    public BPMNSequenceFlowTakenEventHandler(EntityManager entityManager) {
+    public BPMNSequenceFlowTakenEventHandler(
+        EntityManager entityManager,
+        EntityManagerFinder entityManagerFinder
+    ) {
         this.entityManager = entityManager;
+        this.entityManagerFinder = entityManagerFinder;
     }
 
     @Override
     public void handle(CloudRuntimeEvent<?, ?> event) {
         CloudSequenceFlowTakenEvent activityEvent = CloudSequenceFlowTakenEvent.class.cast(event);
+        BPMNSequenceFlow bpmnSequenceFlow = activityEvent.getEntity();
+        var processInstanceId = bpmnSequenceFlow.getProcessInstanceId();
+        var eventId = event.getId();
 
+        entityManagerFinder
+            .findProcessInstanceWithSequenceFlows(processInstanceId)
+            .ifPresent(processInstanceEntity -> {
+                processInstanceEntity
+                    .getSequenceFlowByEventId(eventId)
+                    .ifPresentOrElse(
+                        sequenceFlowEntity -> {
+                            logger.warn(
+                                "Sequence flow '" + sequenceFlowEntity.getElementId()  + "' with eventId '" + eventId +"' already exists in the process '" + processInstanceId + "'!"
+                            );
+                        },
+                        () -> {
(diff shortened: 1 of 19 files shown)


jesty (approved):
LGTM

activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/query/events/handlers/BPMNSequenceFlowTakenEventHandler.java
```diff
@@ -83,11 +72,11 @@ private BPMNSequenceFlowEntity createBpmnSequenceFlowEntity(CloudRuntimeEvent<?,
             event.getAppName(),
             event.getAppVersion()
         );
-        bpmnSequenceFlowEntity.setId(UUID.randomUUID().toString());
+        bpmnSequenceFlowEntity.setId(entityId);
         bpmnSequenceFlowEntity.setElementId(bpmnSequenceFlow.getElementId());
         bpmnSequenceFlowEntity.setProcessDefinitionId(bpmnSequenceFlow.getProcessDefinitionId());
         bpmnSequenceFlowEntity.setProcessInstanceId(bpmnSequenceFlow.getProcessInstanceId());
-        bpmnSequenceFlowEntity.setDate(new Date(activityEvent.getTimestamp()));
+        bpmnSequenceFlowEntity.setDate(new Date(event.getTimestamp()));
```
eromano, line 79: Not sure is correct change the date they can be similar but are actually different, WDYT @erdemedeiros ?

erdemedeiros: Hmm, actually it's not changing the date, it's setting the date for the `bpmnSequenceFlowEntity` created at the line `68` and the value used to set it is the same as before (the variable only get renamed from `activityEvent ` to `event` as part of method extraction.

activiti-cloud-service-common/activiti-cloud-services-test/src/main/java/org/activiti/cloud/starters/test/EventsAggregator.java
```diff
@@ -18,24 +18,59 @@
...
+            .ofNullable(errorMessageRef.get())
+            .map(ErrorMessage.class::cast)
+            .map(ErrorMessage::getPayload)
+            .orElse(null);
+    }
+
+    public EventsAggregator errorChannel(SubscribableChannel errorChannel) {
+        errorChannel.subscribe(this);
+
+        return this;
```
erdemedeiros, line 74: Is this method and all the references to `errorMessageRef` used by tests only? If that's the case, wouldn't it be better to create a dedicated `MessageHandler` in the test scope only?

erdemedeiros (commented):
Great job!
I have one small question concerning the usage of `MessageHandler` in `EventsAggregator`. LGTM, otherwise

activiti-cloud-service-common/activiti-cloud-services-test/src/main/java/org/activiti/cloud/starters/test/EventsAggregator.java
igdianov: It is a test component used always in the test scope. We can refactor this in the next iteration.

andrea-ligios (approved):
LGTM!

activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/query/events/handlers/BPMNSequenceFlowTakenEventHandler.java
```diff
@@ -38,9 +38,32 @@ public BPMNSequenceFlowTakenEventHandler(EntityManager entityManager) {
...
+                        sequenceFlowTakenEvent.getEntity().getProcessInstanceId() +
+                        "'!"
+                    );
+                },
+                () -> {
+                    var sequenceFlowTakenEntity = createBpmnSequenceFlowEntity(entityId, sequenceFlowTakenEvent);
+                    entityManager.persist(sequenceFlowTakenEntity);
+                }
+            );
+    }
```
pmartinezga, line 63: Do we need to do this existence test? I think that once we are using the combination of event id, elemnt id and process id as index, we can skip the checking and use the `entityManager.persist` that is idempotent

PR #1271, opened 2023-11-28
AAE-18550 Reduce memory footprint due dynamic insert overhead in Audit entities

igdianov:
https://alfresco.atlassian.net/browse/AAE-18550


erdemedeiros (approved):
LGTM!

PR #1279, opened 2023-11-30
[AAE-18620] - Fix ZipBuilder test

graric:
The ZipBuilderTest is wrong because the whole zip is compared and not just the contents of the zip.
Debugging I verified and indeed **File last modification date**  bytes are not always the same and this occasionally causes the test to fail.
Rewrote the test using verification on file name and contents.

Bytes headers table

image


atchertchian (approved):

erdemedeiros (approved):
Thank you for getting ride of this random failure!

andrea-ligios (approved):

PR #1291, opened 2023-12-13
AAE-18268 Fix leaking of non-persistent entity attributes into Query Swagger apidocs

igdianov:
This PR fixes Swagger Api generation by whitelisting only persistent JPA attributes  for QuerydslBindings in QuerydslBinderCustomizers used to generate predicate filter names metadata in the Query service ReST Apis.

### DIFF
--- a/activiti-cloud-query-service/activiti-cloud-starter-query/src/test/java/org/activiti/cloud/starter/tests/swagger/QuerySwaggerITSupport.java
+++ b/activiti-cloud-query-service/activiti-cloud-starter-query/src/test/java/org/activiti/cloud/starter/tests/swagger/QuerySwaggerITSupport.java
@@ -50,9 +50,6 @@ public class QuerySwaggerITSupport {
     @Autowired
     private ObjectMapper objectMapper;
 
-    @Value("classpath:swagger-expected.json")
-    private Resource swaggerExpectedResource;
-
     /**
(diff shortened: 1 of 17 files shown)


activiti-cloud-query-service/activiti-cloud-starter-query/src/test/java/org/activiti/cloud/starter/tests/swagger/QuerySwaggerITSupport.java
```diff
@@ -62,5 +68,12 @@ public void generateSwagger() throws Exception {
                 );
                 Files.write(new File("target/swagger.yaml").toPath(), new YAMLMapper().writeValueAsBytes(jsonNodeTree));
             });
+
+        ObjectMapper mapper = new ObjectMapper();
+
+        JsonNode generatedSwagger = mapper.readTree(new File("target/swagger.json"));
+        JsonNode expectedSwagger = mapper.readTree(swaggerExpectedResource.getFile());
+
+        assertThat(generatedSwagger).isEqualTo(expectedSwagger);
```
erdemedeiros: I think it's better moving the assertions to `QuerySwaggerIT` that has already some assertions around the generated swagger and keep `QuerySwaggerITSupport ` for the generation only.

erdemedeiros (approved):
Nice job! I'll just move the assertions from `QuerySwaggerITSupport` to `QuerySwaggerIT` because the first is only suppose to deal with the generation and all the assertions should be handled by the later.

PR #1322, opened 2024-01-12
AAE-19417 latest tag push 

wojciech-piotrowiak:


Makefile
```diff
@@ -92,6 +92,7 @@ docker/%:
 	@echo "Building docker image for $(MODULE):$(RELEASE_VERSION)..."
 	cd activiti-cloud-examples && docker build -f $(MODULE)/Dockerfile -q -t docker.io/activiti/$(MODULE):$(RELEASE_VERSION) $(MODULE)
 	docker push docker.io/activiti/$(MODULE):$(RELEASE_VERSION)
+	docker push docker.io/activiti/$(MODULE):latest
```
atchertchian: this logic is going to be executed on PRs too... it should only push latest tag when run on push on main branch

atchertchian: also shouldn't this be a "docker tag" command instead?

wojciech-piotrowiak: @atchertchian removed second push, and added latest tag to the build command

atchertchian: @wojciech-piotrowiak again this needs to be skipped when run on PRs though: if we can reference explicitly the version used on tests side, that would be cleaner too, as we should not assume latest image is ok for all versions of tests

cc @erdemedeiros

wojciech-piotrowiak: @atchertchian @erdemedeiros 
Agree. Once we will start explicit version in tests, there would be any usecase for 'latest' tag?

atchertchian: We can keep it in case some maintenance branch tests rely on it but I don't see why we would update it anymore indeed

wojciech-piotrowiak: ok, so closing pr and ticket

PR #1350, opened 2024-02-06
AAE-19924 Add label 'be-propagation' to propagation PRs

atchertchian:
This is needed for consistency on labels used for propagations on BE side for AAE-19924, otherwise multiple PRs could be created instead of just one.


erdemedeiros (approved):
LGTM!

PR #1367, opened 2024-02-19
AAE-19764 propagation check of csrf turned on in activiti

wojciech-piotrowiak:


alfresco-build (commented):
Your changes should be placed in modeling project. Project activiti-cloud-modeling-service will be removed soon.

PR #1378, opened 2024-02-23
AAE-20757 Variable type should always be bigdecimal

tom-dal:
Issue link: AAE-20757

PR #1398, opened 2024-03-12
AAE-21097 regression in variable type mapping

tom-dal:
Issue link: AAE-21097

Regression introduced with https://github.com/Activiti/activiti-cloud/pull/1378


jesty (approved):
LGTM

PR #1427, opened 2024-04-23
AAE-21097 Implement permission-based authorization

tom-dal:
Issue link: AAE-22123

### DIFF
--- a/activiti-cloud-service-common/activiti-cloud-services-common-security/src/main/java/org/activiti/cloud/services/common/security/jwt/JwtGrantedAuthorityConverter.java
+++ b/activiti-cloud-service-common/activiti-cloud-services-common-security/src/main/java/org/activiti/cloud/services/common/security/jwt/JwtGrantedAuthorityConverter.java
@@ -17,6 +17,7 @@ package org.activiti.cloud.services.common.security.jwt;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import org.activiti.cloud.services.common.security.CustomAuthorizationManager;
 import org.springframework.core.convert.converter.Converter;
 import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.core.authority.SimpleGrantedAuthority;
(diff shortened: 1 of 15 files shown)


activiti-cloud-service-common/activiti-cloud-services-common-security/src/main/java/org/activiti/cloud/services/common/security/jwt/JwtGrantedAuthorityConverter.java
```diff
@@ -34,8 +34,11 @@ public JwtGrantedAuthorityConverter(JwtAccessTokenProvider jwtAccessTokenProvide
     @Override
     public Collection<GrantedAuthority> convert(Jwt jwt) {
         Collection<GrantedAuthority> grantedAuthorities = new ArrayList<>();
-        for (String authority : jwtAccessTokenProvider.accessToken(jwt).getRoles()) {
-            grantedAuthorities.add(new SimpleGrantedAuthority(this.authorityPrefix + authority));
+        for (String role : jwtAccessTokenProvider.accessToken(jwt).getRoles()) {
+            grantedAuthorities.add(new SimpleGrantedAuthority(this.authorityPrefix + role));
+        }
+        for (String permission : jwtAccessTokenProvider.accessToken(jwt).getPermissions()) {
```
jesty, line 40: should u use PERMISSION_?

jesty (changes requested):

activiti-cloud-service-common/activiti-cloud-services-common-security/src/main/java/org/activiti/cloud/services/common/security/jwt/JwtGrantedAuthorityConverter.java
tom-dal: We can use PERMISSION_ but then we should override default behaviour of Spring which uses ROLE_ for both roles and authorities (there's no concept of permission in the AuthorizationManager).

tom-dal: We can use PERMISSION_ but then we should override default behaviour of Spring which uses ROLE_ for both roles and authorities (there's no concept of permission in the AuthorizationManager).

jesty (approved):
LGTM

PR #1440, opened 2024-05-13
[AAE-22377] Implemented max items limit per request

RCorsaletti:
STILL WP need parameters for max and enabled/disabled

### DIFF
--- a/activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/test/java/org/activiti/cloud/services/query/rest/TaskEntityControllerIT.java
+++ b/activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/test/java/org/activiti/cloud/services/query/rest/TaskEntityControllerIT.java
@@ -252,9 +252,9 @@ class TaskEntityControllerIT {
     }
 
     @Test
-    void testPaginationResolverWithoutSettingMaxOnRequest() throws Exception {
+    void should_returnMaxItemsLimits_when_invokeWithoutPagingParameters() throws Exception {
         //given
-        AlfrescoPageRequest pageRequest = new AlfrescoPageRequest(1001, 1000, PageRequest.of(0, 1000));
+        AlfrescoPageRequest pageRequest = new AlfrescoPageRequest(0, 1000, PageRequest.of(0, 1000));
 
         given(taskRepository.findAll(any(), eq(pageRequest)))
             .willReturn(new PageImpl<>(Collections.singletonList(buildDefaultTask()), pageRequest, 1000));
@@ -268,7 +268,7 @@ class TaskEntityControllerIT {
 
         assertThatJson(result.getResponse().getContentAsString())
             .node("list.pagination.skipCount")
-            .isEqualTo(1001)
+            .isEqualTo(0)
             .node("list.pagination.maxItems")
             .isEqualTo(1000)
             .node("list.pagination.count")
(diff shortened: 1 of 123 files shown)


activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/main/java/org/activiti/cloud/alfresco/argument/resolver/AlfrescoPageArgumentMethodResolver.java
```diff
@@ -29,6 +30,9 @@ public class AlfrescoPageArgumentMethodResolver implements PageableArgumentResol
     private final AlfrescoPageParameterParser pageParameterParser;
     private final PageableHandlerMethodArgumentResolver pageableHandlerMethodArgumentResolver;
 
+    private static final int MAX_ITEMS_LIMIT = 1000; // add parameter
+    private static final boolean MAX_ITEMS_LIMIT_ENABLED = true; // add parameter
```
graric: Are these comments needed?

activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/main/java/org/activiti/cloud/alfresco/argument/resolver/AlfrescoPageArgumentMethodResolver.java
```diff
@@ -58,6 +62,15 @@ public Pageable resolveArgument(
         );
 
         AlfrescoQueryParameters alfrescoQueryParameters = pageParameterParser.parseParameters(webRequest);
+        boolean isMaxItemsExceedingLimit =
+            !alfrescoQueryParameters.getMaxItemsParameter().isSet() ||
+            alfrescoQueryParameters.getMaxItemsParameter().getValue() > MAX_ITEMS_LIMIT;
+
+        if (isMaxItemsExceedingLimit && MAX_ITEMS_LIMIT_ENABLED) {
```
graric: What if we move these lines into a method and just call it directly in the if? 
That way we can make it more readable IMO

activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/test/java/org/activiti/cloud/services/query/rest/TaskEntityControllerIT.java
```diff
@@ -114,7 +114,7 @@ void setUp() {
     }
 
     @Test
-    public void findAllShouldReturnAllResultsUsingAlfrescoMetadataWhenMediaTypeIsApplicationJson() throws Exception {
+    void findAllShouldReturnAllResultsUsingAlfrescoMetadataWhenMediaTypeIsApplicationJson() throws Exception {
```
graric, line 123: Can you please follow the test naming convention? 🙏 

https://hyland.atlassian.net/wiki/spaces/HXP/pages/1308014107/Studio+-+Coding+Style+Guidelines#Naming-and-other-test-conventions

graric: Also, the public access modifier is no longer needed in JUnit 5 tests, which is great news!

activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/main/java/org/activiti/cloud/alfresco/argument/resolver/AlfrescoPageArgumentMethodResolver.java
erdemedeiros: I think both values should use system properties, so that they can be overridden if needed. I'd also use `false` as default value for `MAX_ITEMS_LIMIT_ENABLED`.

activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/main/java/org/activiti/cloud/alfresco/argument/resolver/AlfrescoPageArgumentMethodResolver.java
erdemedeiros: +1

activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/main/java/org/activiti/cloud/alfresco/argument/resolver/AlfrescoPageArgumentMethodResolver.java
```diff
@@ -58,6 +62,15 @@ public Pageable resolveArgument(
...
+            Pageable limitedPageable = PageRequest.of(0, MAX_ITEMS_LIMIT);
+            return new AlfrescoPageRequest(MAX_ITEMS_LIMIT + 1L, MAX_ITEMS_LIMIT, limitedPageable);
```
erdemedeiros: IMO, silently changing the values provided via query param sounds confusing to me and I think people will report bugs for this. Maybe it's better to raise an exception in this case.

activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/test/java/org/activiti/cloud/services/query/rest/TaskEntityControllerIT.java
```diff
@@ -250,4 +250,60 @@ public void should_returnTaskPermissions_when_invokeGetTaskById() throws Excepti
...
+            .isEqualTo(1001)
+            .node("list.pagination.maxItems")
+            .isEqualTo(1000)
+            .node("list.pagination.count")
+            .isEqualTo(1)
+            .node("list.pagination.hasMoreItems")
+            .isEqualTo(false)
+            .node("list.pagination.totalItems")
+            .isEqualTo(1002);
+    }
```
erdemedeiros: In any case, we should never touch to `skipCount` that gives how many elements should be left behind before getting the first element to be returned (this is the equivalent of start index). In this specific case, the caller is asking for up to `2000` elements, skipping the first 2001 elements from the results, but what he's getting in return is up `1000` (what could still be fine), skipping only the first 1001 elements from the results: he's getting 1000 elements that he's not expecting.

erdemedeiros (changes requested):
Hello, I think this change is going to be confusing for end users. Let's discuss about it.

PR #1449, opened 2024-05-24
AAE-22377 Exception if user exceed pagination limit

RCorsaletti:
https://hyland.atlassian.net/browse/AAE-22377


activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/main/java/org/activiti/cloud/alfresco/argument/resolver/AlfrescoPageArgumentMethodResolver.java
```diff
@@ -71,4 +81,33 @@ public Pageable resolveArgument(
...
+                alfrescoQueryParameters.getMaxItemsParameter().getValue() > maxItemsLimit
+            ) {
+                return true;
+            }
+            if (
+                alfrescoQueryParameters.getPageParameter().isSet() &&
+                alfrescoQueryParameters.getPageParameter().getValue() > maxItemsLimit
+            ) {
+                return true;
+            } else if (
```
jsokolowskii:
```suggestion
            } else if (!alfrescoQueryParameters.getMaxItemsParameter().isSet() && basePageable.getPageSize() > maxItemsLimit) {
            return true;
            }
```
I think this should work and we could remove the `return true;` at the end. WDYT?

activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/test/java/org/activiti/cloud/alfresco/argument/resolver/AlfrescoPageArgumentMethodResolverTest.java
```diff
@@ -37,7 +37,7 @@
 @ExtendWith(MockitoExtension.class)
 public class AlfrescoPageArgumentMethodResolverTest {
 
-    @InjectMocks
+    //    @InjectMocks
```
jsokolowskii: Commented code

activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/main/resources/config/alfresco-rest-config.properties
```diff
@@ -1,3 +1,6 @@
 spring.data.rest.default-page-size=100
 spring.data.rest.default-media-type=application/json
 spring.hateoas.use-hal-as-default-json-media-type=false
+
+max_items_limit=${MAX_ITEMS_LIMIT:1000}
+max_items_limit_enabled=${MAX_ITEMS_LIMIT_ENABLED:true}
```
atchertchian: Is this extra property needed? limit=0 could mean "no limit"

RCorsaletti: we decided to add a limit to the maxItems parameter to 1000 elements, but if there is the need to tweak that i also added an env variable to set a different limit. 0 would be 0 items limit in this case so no items will be retrieved, to disable the maxLimit there is the MAX_ITEMS_LIMIT_ENABLED, if set to false it will disable the check.

atchertchian: I understood that, I mean that limit=0 could be handled in a specific way (eg no limit) because there are no use cases where "0 items" configuration would be useful/meaningful anyway.
But that's kind of a detail, I don't see the point of this extra property that's all.

activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/main/java/org/activiti/cloud/alfresco/config/AlfrescoWebAutoConfiguration.java
```diff
@@ -50,13 +50,19 @@ public class AlfrescoWebAutoConfiguration implements WebMvcConfigurer {
 
     private final PageableHandlerMethodArgumentResolver pageableHandlerMethodArgumentResolver;
     private final int defaultPageSize;
+    private final int maxItemsLimit;
+    private final boolean maxItemsLimitEnabled;
 
     public AlfrescoWebAutoConfiguration(
         @Lazy PageableHandlerMethodArgumentResolver pageableHandlerMethodArgumentResolver,
-        @Value("${spring.data.rest.default-page-size:100}") int defaultPageSize
+        @Value("${spring.data.rest.default-page-size:100}") int defaultPageSize,
```
jesty, line 58: why default it's 100 if usually in UI we show 25 results?

jesty (changes requested):

activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/main/java/org/activiti/cloud/alfresco/config/AlfrescoWebAutoConfiguration.java
RCorsaletti: As we agreed i'll change the default to 25 results

jesty (approved):
LGTM

activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/test/java/org/activiti/cloud/services/query/rest/TaskEntityControllerIT.java
```diff
@@ -250,4 +255,32 @@ public void should_returnTaskPermissions_when_invokeGetTaskById() throws Excepti
             .ofLength(1)
             .thatContains(TaskPermissions.VIEW);
     }
+
+    @Test
+    void should_returnMaxItemsLimits_when_invokeWithPagingParametersExceedingLimits() throws Exception {
```
erdemedeiros:
```suggestion
    void should_returnBadRequest_when_invokeWithPagingParametersExceedingLimits() throws Exception {
```

activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/test/java/org/activiti/cloud/services/query/rest/TaskEntityControllerIT.java
```diff
@@ -250,4 +255,32 @@ public void should_returnTaskPermissions_when_invokeGetTaskById() throws Excepti
...
+        //given
+        AlfrescoPageRequest pageRequest = new AlfrescoPageRequest(1000, 1000, PageRequest.of(0, 1000));
+
+        given(taskRepository.findAll(any(), eq(pageRequest)))
+            .willReturn(new PageImpl<>(Collections.singletonList(buildDefaultTask()), pageRequest, 2000));
+
+        //when
+        mockMvc
+            .perform(get("/v1/tasks?skipCount=1000&maxItems=2000").accept(MediaType.APPLICATION_JSON))
+            .andExpect(status().isBadRequest());
```
erdemedeiros: In addition, to the status, I think it would be nice to assert that the message contains the explanation why it's a bad request

activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/main/java/org/activiti/cloud/alfresco/argument/resolver/AlfrescoPageArgumentMethodResolver.java
```diff
@@ -58,7 +65,10 @@ public Pageable resolveArgument(
         );
 
         AlfrescoQueryParameters alfrescoQueryParameters = pageParameterParser.parseParameters(webRequest);
-        if (
+
+        if (isPaginationValueExceedinglimit(alfrescoQueryParameters, basePageable)) {
+            throw new IllegalStateException("Exceeded max limit of 1000 elements");
```
erdemedeiros: It should reuse the field `maxItemsLimit` instead of hard coded `1000`

activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/main/java/org/activiti/cloud/alfresco/argument/resolver/AlfrescoPageArgumentMethodResolver.java
```diff
@@ -71,4 +81,32 @@ public Pageable resolveArgument(
             return basePageable;
         }
     }
+
+    private boolean isPaginationValueExceedinglimit(
```
erdemedeiros: Missing capital letter on `limit` : `isPaginationValueExceedinglimit` -> `isPaginationValueExceedingLimit`

activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/test/java/org/activiti/cloud/services/query/rest/TaskEntityControllerIT.java
```diff
@@ -250,4 +255,32 @@ public void should_returnTaskPermissions_when_invokeGetTaskById() throws Excepti
...
+    }
+
+    @Test
+    void should_returnMaxItemsLimits_when_invokeWithPageParameterExceedingLimits() throws Exception {
```
erdemedeiros: Maybe we can have a single test here, just to cover the conversion from java exception to BadRequest and then have more fine grain test were we can cover the limits using unit tests in `AlfrescoPageArgumentMethodResolverTest`. The tests will run faster there because they don't need to load spring context. If it helps you could even make the method `isPaginationValueExceedingLimit` protected or public to test directly it's output. WDYT?

activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/main/java/org/activiti/cloud/alfresco/argument/resolver/AlfrescoPageParameterParser.java
```diff
@@ -50,4 +54,16 @@ protected SkipCountParameter parseSkipCount(NativeWebRequest webRequest) {
...
+        int maxItems = defaultPageSize;
+
+        String maxItemsString = webRequest.getParameter("size");
+        boolean isSet = maxItemsString != null;
+        if (isSet) {
+            maxItems = Integer.parseInt(maxItemsString);
+        }
+
+        return new PageSizeParameter(isSet, maxItems);
+    }
```
erdemedeiros: I don't think you need to add the logic to parse `size` as it's already handled by Spring and it will be available via `basePageable.getPageSize()`.

activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/main/java/org/activiti/cloud/alfresco/argument/resolver/AlfrescoPageArgumentMethodResolver.java
```diff
@@ -71,4 +81,32 @@ public Pageable resolveArgument(
...
+            } else if (
+                alfrescoQueryParameters.getMaxItemsParameter().isSet() &&
+                alfrescoQueryParameters.getMaxItemsParameter().getValue() < maxItemsLimit
+            ) {
+                return false;
+            } else {
+                return basePageable.getPageSize() >= maxItemsLimit;
+            }
+        }
+        return false;
```
erdemedeiros: I think this entire block can be simplified to the code bellow, because there are only two cases that are problematic:
1. when the Alfresco naming convention is used (`maxItems` set): maxItems > maxItemsLimit
2. when Spring naming conversion is used (`maxItems` no set): pageSize > maxItemsLimit
```suggestion
        if (maxItemsLimitEnabled) {
            if (alfrescoQueryParameters.getMaxItemsParameter().isSet()) {
                return alfrescoQueryParameters.getMaxItemsParameter().getValue() > maxItemsLimit;
            } else {
                return basePageable.getPageSize() > maxItemsLimit;
            }
        }
        return false;
```

activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/main/java/org/activiti/cloud/alfresco/argument/resolver/AlfrescoQueryParameters.java
```diff
@@ -21,9 +21,16 @@ public class AlfrescoQueryParameters {
 
     private MaxItemsParameter maxItemsParameter;
 
-    public AlfrescoQueryParameters(SkipCountParameter skipCountParameter, MaxItemsParameter maxItemsParameter) {
+    private PageSizeParameter pageSizeParameter;
```
erdemedeiros: Not needed.

activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/main/java/org/activiti/cloud/alfresco/argument/resolver/PageSizeParameter.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.activiti.cloud.alfresco.argument.resolver;
+
+public class PageSizeParameter extends AlfrescoParameter<Integer> {
```
erdemedeiros: Not needed.

activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/main/java/org/activiti/cloud/alfresco/config/AlfrescoWebAutoConfiguration.java
```diff
@@ -50,13 +50,19 @@ public class AlfrescoWebAutoConfiguration implements WebMvcConfigurer {
...
+        @Value("${max_items_limit:1000}") int maxItemsLimit,
+        @Value("${max_items_limit_enabled:false}") boolean maxItemsLimitEnabled
```
erdemedeiros: I'd suggest to rename the properties to use dots and hyphens as separators (the separators most used in the current properties) and to add  prefix to give more context and avoid potential conflicts. For instance:
- `activiti.cloud.rest.max-items`
- `activiti.cloud.rest.maxt-items.enabled`

activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/test/java/org/activiti/cloud/alfresco/argument/resolver/AlfrescoPageArgumentMethodResolverTest.java
```diff
@@ -75,81 +79,4 @@ private MethodParameter buildParameter(Class<?> parameterType) {
         doReturn(parameterType).when(parameter).getParameterType();
         return parameter;
     }
-
-    @Test
-    public void resolveArgumentShouldReturnAPageableBasedOnParsedSkipCountAndMaxItems() throws Exception {
```
erdemedeiros: Why removing these two tests?

erdemedeiros (changes requested):
I added some suggestions. Just to give some context about the different naming used: for historical reasons there are two flavors of naming conventions that are accepted in order to get paged results from the REST API: Alfresco naming convention and Spring naming convention.
- Alfresco: the pagination is controlled via `skipCount` and `maxItems` query parameters. `skipCount` defines how many elements should be skipped before the first one to be included in the results and `maxItems` defines the maximum number of elements that may be included for the given request (i.e, the page size)
- Spring: the pagination is controlled via `page` and `size` query parameters. `page` defines the number of the page and `size` the maximum number of elements that may be included in the given page
As an example, `page=2&size=10` in Spring will be equivalent of `skipCount=20&size=10` as we are skipping the elements as we are getting the page `2` and so skipping the elements of pages `0` and `1`

activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/test/java/org/activiti/cloud/alfresco/argument/resolver/AlfrescoPageArgumentMethodResolverTest.java
RCorsaletti: Fixed the mock of AlfrescoPageArgumentMethodResolver and uncommented the tests

erdemedeiros (approved):
LGTM!

PR #1492, opened 2024-07-30
AAE-22454 Process Task list is not showing tasks when process variables are added to the Column process variables

tom-dal:
Issue link: https://github.com/Activiti/Activiti/issues/4729


activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-model/pom.xml
```diff
@@ -118,6 +118,11 @@
       <artifactId>jackson-datatype-jsr310</artifactId>
       <version>2.16.1</version>
     </dependency>
+    <dependency>
+      <groupId>org.hibernate</groupId>
+      <artifactId>hibernate-jpamodelgen</artifactId>
```
mavotto, line 123: This dependency should enable the metamodel creation, but I don't see it being used in your changes. Is it a leftover?

tom-dal: It's being used in ProcessVariableSpecification, but I forgot to use it in the last filter. Thanks!

mavotto: I miss the part where it was used, and I only see the one where it was missing 😄

tom-dal: Lines 45-46

mavotto (approved):

activiti-cloud-query-service/activiti-cloud-starter-query/src/test/java/org/activiti/cloud/starter/tests/QueryTasksIT.java
```diff
@@ -2651,21 +2651,6 @@ void should_migrateTaskProcessVariables() throws Exception {
         taskProcessVariableCount = processVariablesMigrationHelper.getTaskProcessVariableCount(task.getId());
         assertThat(taskProcessVariableCount).isEqualTo(0);
 
-        await()
```
jesty, line 2654: does this test cover the bug? if not, can we add a test for it? thanks!

tom-dal: This one is the test for the task migration script that was failing because it was deleting associations from task_process_variable table and then using the new approach to check the removal, but the new approach does not read from that table. I think the test is fine even without checking if the associations have been removed from the table (we're doing it with a native sql query, it shouldn't be purpose of it to test a native sql query).

For the bug, I added this test which was failing with the old approach:
https://github.com/Activiti/activiti-cloud/blob/0c2454ad2e77fa2005c2a88173e873030b35833b/activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/test/java/org/activiti/cloud/services/query/rest/TaskControllerHelperIT.java#L215

PR #1534, opened 2024-09-10
AAE-25834 Fix start form constants when no form

pmartinezga:
When there is no form key in the start event, the constants for it should be retrieved anyway


jesty (approved):
LGTM

jesty (approved):
LGTM

mavotto (approved):

PR #1537, opened 2024-09-10
AAE-24339 Refactoring to reuse code for process instance filters

tom-dal:
AAE-24339


activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/query/rest/payload/TaskSearchRequest.java
```diff
@@ -16,272 +16,11 @@
...
-
-    public Set<ProcessVariableKey> getProcessVariableKeys() {
-        return processVariableKeys;
-    }
-
-    public void setProcessVariableKeys(Set<ProcessVariableKey> processVariableKeys) {
-        this.processVariableKeys = processVariableKeys;
-    }
-}
+public record TaskSearchRequest(
```
jesty, line 25: can we have a field for each line? Thanks!

activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/query/rest/specification/SpecificationSupport.java
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    protected Predicate getVariableValueCondition(
+        Path<?> valueColumnPath,
+        VariableFilter filter,
+        CriteriaBuilder criteriaBuilder
+    ) {
+        return criteriaBuilder.isTrue(
+            switch (filter.operator()) {
+                case EQUALS -> switch (filter.type()) {
```
jesty: instead of having a case of cases, can we replace something like a strategy? Where we have a class with one strategy for each type (booelan, integer, ...), of course an interface, and the stratehy selector here. Thanks!

jesty (changes requested):
Few changes, where are tests? thanks!

jesty (approved):
LGTM

PR #1542, opened 2024-09-17
AAE-25874 Drop index on process variable when field type changes from text to jsonb

tom-dal:
AAE-25874


gicappa (commented):
Modifying an existing liquibase script already deployed is generally speaking a bad idea: the environment that have this script already launched would break. 
I would be very careful with this change and only if you are 100% sure this change has not been applied anywhere we can go with it.

eromano (approved):

mavotto (approved):

tom-dal:
Tested retrocompatibility with already executed scripts by running the new image on running applications.

Tested on application with manually created index by following these steps:

- deployed an application
- dropped and recreated application schema
- re-initialized application schema by mounting query-liquibase 7.15.0
- created index manually
- mounted new query-liquibase-image
- checked that no errors occurred and index has been removed

Also tested on application with query service 7.15.0 with no index created.

PR #1543, opened 2024-09-18
AAE-22831 Fix status is changed to 'Cancelled' when deleting process instance

igdianov:
The client is chaining two api calls for Process Instance DELETE which cancels the process followed by DESTROY, which sends a PROCESS_DELETED event to Query servicd.

After that, the client is trying to refresh the list of the processes, and if the DESTROY event is lagging, so it will get an intermediate CANCELLED status, but after refreshing the list again, the process will disappear from the Query database.

https://hyland.atlassian.net/browse/AAE-22831

This PR extends the destroy process instance REST api call to actually cancel the process execution in the Runtime in one transaction which will send the delete process instance cloud event to the Query service in one message. 

The extended Runtime Admin API call will support a new `force=true` parameter to be able to destroy the process in one command:

```curl
DELETE /admin/v1/process-instances/{id}/destroy?force=true
```

The existing destroy process api command without `force` parameter will work as before by declining to destroy if the status of the process is not CANCELLED or COMPLETED.

### DIFF
--- a/activiti-cloud-audit-service/activiti-cloud-services-audit-model/src/test/java/org/activiti/cloud/services/audit/jpa/converters/ProcessDeletedEventConverterTest.java
+++ b/activiti-cloud-audit-service/activiti-cloud-services-audit-model/src/test/java/org/activiti/cloud/services/audit/jpa/converters/ProcessDeletedEventConverterTest.java
@@ -16,8 +16,6 @@
 package org.activiti.cloud.services.audit.jpa.converters;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.BDDMockito.given;
 
 import java.util.Arrays;
 import java.util.List;
@@ -37,7 +35,6 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
-import org.springframework.data.domain.Sort;
 
 @ExtendWith(MockitoExtension.class)
 public class ProcessDeletedEventConverterTest {
@@ -65,10 +62,8 @@ public class ProcessDeletedEventConverterTest {
         //given
         String processDefinitionId = UUID.randomUUID().toString();
         String processInstanceId = UUID.randomUUID().toString();
-        given(eventsRepository.findAll(any(), any(Sort.class)))
-            .willReturn(buildCompletedEntities(processDefinitionId, processInstanceId));
 
-        CloudRuntimeEvent<?, ?> runtimeEvent = buildEvent(processInstanceId);
+        CloudRuntimeEvent<?, ?> runtimeEvent = buildEvent(processInstanceId, processDefinitionId);
 
         //when
         ProcessDeletedAuditEventEntity event = converter.createEventEntity(runtimeEvent);
@@ -83,10 +78,8 @@ public class ProcessDeletedEventConverterTest {
         //given
         String processDefinitionId = UUID.randomUUID().toString();
         String processInstanceId = UUID.randomUUID().toString();
-        given(eventsRepository.findAll(any(), any(Sort.class)))
-            .willReturn(buildCompletedEntities(processDefinitionId, processInstanceId));
 
-        CloudRuntimeEvent<?, ?> runtimeEvent = buildEvent(processInstanceId);
+        CloudRuntimeEvent<?, ?> runtimeEvent = buildEvent(processInstanceId, processDefinitionId);
 
         //when
         ProcessDeletedAuditEventEntity event = converter.createEventEntity(runtimeEvent);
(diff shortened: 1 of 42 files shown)


atchertchian (approved):

activiti-cloud-audit-service/activiti-cloud-services-audit-model/src/test/java/org/activiti/cloud/services/audit/jpa/converters/ProcessDeletedEventConverterTest.java
```diff
@@ -97,9 +90,10 @@ public void createAPIEventShouldReturnEvent() {
         assertThat(apiEvent.getProcessInstanceId()).isEqualTo(processInstanceId);
     }
 
-    private CloudRuntimeEvent<?, ?> buildEvent(String processInstanceId) {
+    private CloudRuntimeEvent<?, ?> buildEvent(String processInstanceId, String processDefinitionId) {
         ProcessInstanceImpl instance = new ProcessInstanceImpl();
         instance.setId(processInstanceId);
+        instance.setProcessDefinitionId(processDefinitionId);
```
erdemedeiros, line 96: Ins't processDefinitionId going to be missing in the real event? In this case better to remove it from here as well.

activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-rest/src/main/java/org/activiti/cloud/services/query/rest/ProcessInstanceAdminService.java
```diff
@@ -102,7 +102,7 @@ public ProcessInstanceEntity findById(@PathVariable String processInstanceId) {
         return entityFinder.findById(
             processInstanceRepository,
             processInstanceId,
-            "Unable to find task for the given id:'" + processInstanceId + "'"
+            "Unable to find process instance for the given id:'" + processInstanceId + "'"
```
erdemedeiros, line 110: Good catch!

activiti-cloud-query-service/activiti-cloud-starter-query/src/test/java/org/activiti/cloud/starter/tests/QueryProcessInstancesEntityIT.java
```diff
@@ -261,6 +263,52 @@ public void shouldGetAdminProcessInfo() {
...
+                identityTokenProducer.withTestUser("hradmin");
+
+                ResponseEntity<ProcessInstance> responseEntity = testRestTemplate.exchange(
+                    ADMIN_PROC_URL + "/" + process.getId(),
+                    HttpMethod.GET,
+                    identityTokenProducer.entityWithAuthorizationHeader(),
+                    new ParameterizedTypeReference<>() {}
+                );
+                //then
+                assertThat(responseEntity.getStatusCode()).isEqualTo(HttpStatus.NOT_FOUND);
```
erdemedeiros, line 349: Shouldn't we add coverage for related entities as well (tasks, candidate groups, etc)?

activiti-cloud-runtime-bundle-service/activiti-cloud-starter-runtime-bundle/src/test/java/org/activiti/cloud/starter/tests/runtime/ProcessInstanceIT.java
```diff
@@ -666,6 +672,40 @@ void adminShouldDeleteProcessInstance() {
...
+        ResponseEntity<CloudProcessInstance> responseEntity = processInstanceRestTemplate.adminDestroy(
+            processEntity,
+            true
+        );
+
+        //then
+        assertThat(responseEntity.getStatusCode()).isEqualTo(HttpStatus.OK);
+
+        assertThatExceptionOfType(AssertionError.class)
+            .isThrownBy(() -> processInstanceRestTemplate.getProcessInstance(processEntity));
```
erdemedeiros: This assertion is a bit weird as it's asserting that there is an assertion error. I guess you are running in this case because there is assertion that is failing inside `getProcessInstance` where the status code is `NOT_FOUND` instead of `OK`. IMO, it's better to rewrite this to make assertion on status code here.

activiti-cloud-audit-service/activiti-cloud-services-audit-model/src/test/java/org/activiti/cloud/services/audit/jpa/converters/ProcessDeletedEventConverterTest.java
igdianov: This is OK, because `SendDeleteCloudProcessInstanceEventCmd` will use `ProcessEngineEventsAggregator` to append process definition attributes from the execution context available inside transaction.

activiti-cloud-runtime-bundle-service/activiti-cloud-starter-runtime-bundle/src/test/java/org/activiti/cloud/starter/tests/runtime/ProcessInstanceIT.java
igdianov: This is because `processInstanceRestTemplate.getProcessInstance` method call will throw assertion exception if the response is no `200 OK`. I extended the assert to check for `404 NOT FOUND` error message.

activiti-cloud-query-service/activiti-cloud-starter-query/src/test/java/org/activiti/cloud/starter/tests/QueryProcessInstancesEntityIT.java
igdianov: I added test coverage for the related entities when deleting a process instance.

erdemedeiros (approved):
LGTM

PR #1544, opened 2024-09-19
AAE-23154 Fix range filters

tom-dal:
AAE-23154

Range filters were not working correctly.
For range filter is intended the simultaneous use of gt and lt operators.


jesty (approved):
LGTM

PR #1547, opened 2024-09-20
AAE-25874 Create indexes on task and task_process_variable

### LINKED ISSUE #4786: Slow query on task and task_process_variable
`select te1_0.id from task te1_0 left join task_process_variable pv1_0 on te1_0.id = pv1_0.task_id left join process_variable pv1_1 on pv1_1.id = pv1_0.process_variable_id left join task_process_variable pv2_0 on te1_0.id = pv2_0.task_id left join process_variable pv2_1 on pv2_1.id = pv2_0.process_variable_id where pv1_1.name = ? and pv1_1. value = ? and pv2_1.name = ? and pv2_1. value = ? and te1_0.name like ? escape ? and te1_0.status in ( ? ) order by te1_0.created_date offset ? rows fetch first ? rows on`

AAE-25874

tom-dal:
AAE-25874

### DIFF
--- a/activiti-cloud-api/pom.xml
+++ b/activiti-cloud-api/pom.xml
@@ -12,7 +12,7 @@
   <name>Activiti Cloud :: Runtime API Parent</name>
   <packaging>pom</packaging>
   <properties>
-    <activiti.version>8.7.0-alpha.22</activiti.version>
+    <activiti.version>8.7.0-alpha.23</activiti.version>
   </properties>
   <modules>
     <module>activiti-cloud-api-dependencies</module>
--- a/activiti-cloud-api/pom.xml
+++ b/activiti-cloud-api/pom.xml
@@ -12,7 +12,7 @@
   <name>Activiti Cloud :: Runtime API Parent</name>
   <packaging>pom</packaging>
   <properties>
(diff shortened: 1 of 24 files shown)


jesty (commented):
In the issue I saw a index on value field on process_variable table. Why we don't have here?

tom-dal:
@jesty the index on process_variable in the issue is the one that we created manually on the old version of activiti where value is a string field. We took care of removing this index before changing value column type with this changeset:
https://github.com/Activiti/activiti-cloud/blob/b2c1f587a5392e8e18b73103727a00c68841e5f8/activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-liquibase/src/main/resources/config/query/liquibase/master.xml#L495

 With the new schema where the value is a jsonb we have some indexes here:
https://github.com/Activiti/activiti-cloud/blob/b2c1f587a5392e8e18b73103727a00c68841e5f8/activiti-cloud-query-service/activiti-cloud-services-query/activiti-cloud-services-query-liquibase/src/main/resources/config/query/liquibase/changelog/24-alter.pg.schema.8.7.0.sql#L23

jesty (approved):
LGTM

PR #1548, opened 2024-09-20
AAE-26067 Spring boot upgrade to 3.3.4

wojciech-piotrowiak:


activiti-cloud-service-common/activiti-cloud-service-messaging-config/src/main/java/org/activiti/cloud/common/messaging/config/FunctionBindingConfiguration.java
```diff
@@ -188,4 +198,24 @@ public Object postProcessAfterInitialization(Object bean, String beanName) throw
...
     public interface BindingResolver extends Function<String, String> {}
+
+    @Bean
+    @ConditionalOnClass(JacksonMapper.class)
+    @Primary
+    public JacksonMapper jacksonMapper(@Autowired(required = false) ObjectMapper objectMapper) {
+        //temporary workaround for https://github.com/spring-cloud/spring-cloud-function/issues/1159
+        ObjectMapper copiedMapper;
+        try {
+            copiedMapper = objectMapper.copy();
```
erdemedeiros: As `objectMapper` is optional, it's better check if it's `null` before calling `copy` method to avoid potential NPE. If it's null you can already instantiate a new one.

jsokolowskii (approved):

activiti-cloud-service-common/activiti-cloud-service-messaging-config/src/main/java/org/activiti/cloud/common/messaging/config/FunctionBindingConfiguration.java
```diff
@@ -188,4 +198,24 @@ public Object postProcessAfterInitialization(Object bean, String beanName) throw
...
+            //logic from AlfrescoWebAutoConfiguration.configureObjectMapperForBigDecimal
+            copiedMapper.configOverride(BigDecimal.class).setFormat(JsonFormat.Value.forShape(JsonFormat.Shape.STRING));
+        } catch (Exception e) {
+            copiedMapper = new ObjectMapper();
+            copiedMapper.registerModule(new JavaTimeModule());
+            copiedMapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
+            copiedMapper.configure(DeserializationFeature.FAIL_ON_TRAILING_TOKENS, true);
+            copiedMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
erdemedeiros, line 213: Is it expected that this is called only in the case of exception? Shouldn't we always go through it?

activiti-cloud-service-common/activiti-cloud-service-messaging-config/src/main/java/org/activiti/cloud/common/messaging/config/FunctionBindingConfiguration.java
```diff
@@ -188,4 +198,24 @@ public Object postProcessAfterInitialization(Object bean, String beanName) throw
...
+    @Bean
+    @ConditionalOnClass(JacksonMapper.class)
+    @Primary
+    public JacksonMapper jacksonMapper(@Autowired(required = false) ObjectMapper objectMapper) {
+        //temporary workaround for https://github.com/spring-cloud/spring-cloud-function/issues/1159
+        ObjectMapper copiedMapper;
+        try {
+            copiedMapper = objectMapper.copy();
+            //logic from AlfrescoWebAutoConfiguration.configureObjectMapperForBigDecimal
+            copiedMapper.configOverride(BigDecimal.class).setFormat(JsonFormat.Value.forShape(JsonFormat.Shape.STRING));
```
erdemedeiros: Shouldn't this be called also in the case where a new Object mapper is instantiated (catch exception block)?

activiti-cloud-service-common/activiti-cloud-service-messaging-config/src/main/java/org/activiti/cloud/common/messaging/config/FunctionBindingConfiguration.java
wojciech-piotrowiak: NPE will be handled in the same way as an issue with copying existing objectMapper that's why I did not checked it for null

activiti-cloud-service-common/activiti-cloud-service-messaging-config/src/main/java/org/activiti/cloud/common/messaging/config/FunctionBindingConfiguration.java
wojciech-piotrowiak: agree, moved

activiti-cloud-service-common/activiti-cloud-service-messaging-config/src/main/java/org/activiti/cloud/common/messaging/config/FunctionBindingConfiguration.java
wojciech-piotrowiak: Ok, moved those settings outside of the catch. There is no risk since we modify copied object not a common one from the context

activiti-cloud-service-common/activiti-cloud-service-messaging-config/src/main/java/org/activiti/cloud/common/messaging/config/FunctionBindingConfiguration.java
erdemedeiros: Yes, I know that the case of NPE is already covered by the generic catch exception, but usually using exceptions as control flow is considered as an anti-pattern and the check for null is easy to put in place, so I'd prefer avoiding NPE, rather than throwing + catching the NPE.

erdemedeiros (approved):

PR #1550, opened 2024-09-20
AAE-25958 Fix Query Service OOM errors caused by infinite/unbounded service tasks instances

igdianov:
https://hyland.atlassian.net/browse/AAE-25958


erdemedeiros (approved):
LGTM!

PR #1562, opened 2024-10-02
AAE-25874 Test improvements for ProcessInstance filters

tom-dal:
AAE-25874

ProcessInstance filters are now tested directly via WebMVC layer by invoking both ProcessInstanceController and ProcessInstanceAdminController (also testing the process list restriction on the first controller).


jesty (approved):
LGTM

PR #1569, opened 2024-10-11
AAE-26470 Get less user data from keycloak in order to speedup the response

jesty:
Keycloak 22 it's quite slow when searching for users, so we speedup a bit the call passing a parameter in order to have less user details.


tom-dal (approved):
LGTM

PR #1571, opened 2024-10-11
AAE-25188 Task search test improvements and refactoring

tom-dal:
AAE-25188

Task filters are now tested directly via WebMVC layer by invoking both TaskController and TaskAdminController (also testing the task list restriction on the first controller).


jesty (approved):
LGTM

PR #1582, opened 2024-10-21
AAE-26100 Synchronize process definitions between runtime bundle and query services

igdianov:
This is part 1 of the implementation to add a command executor in the Process Runtime Service core to be able to trigger process definitions sync via a command message sent by the Query service via a batch job. The Process Runtime will handle the synchronization request by sending the process definitions back to Query via existing process deployed audit messages and return execution result to the requestor via command results channel.

Part of https://hyland.atlassian.net/browse/AAE-26100


activiti-cloud-acceptance-tests/activiti-cloud-acceptance-tests-core/src/main/java/org/activiti/cloud/acc/core/steps/query/ProcessQuerySteps.java
```diff
@@ -112,6 +113,7 @@ public void checkProcessInstanceHasVariableValue(
         Object variableValue
     ) {
         await()
+            .atMost(Duration.ofSeconds(30))
```
erdemedeiros: Is this needed? I was expecting the default timeout of 10 seconds to be enough.

erdemedeiros (approved):
Great job @igdianov ! I like the usage of IntegrationFlows with the auto routing based on enriched headers. Maybe we can evolve the connector starters to use something similar in the future. 
I've added only one small question. LGTM, otherwise

PR #1619, opened 2024-11-25
AAE-28598 Add "not equals" process variables filter for task and process instance search

### LINKED ISSUE #4853: Add "not equals" process variables filter for task and process instance search
AAE-28598

tom-dal:
Issue link: AAE-28598


jesty (approved):
LGTM

PR #1658, opened 2025-01-10
AAE-29955 Fix bigdecimal filtering precision issues

### LINKED ISSUE #4890: Bigdecimal variable filtering gives wrong results due to scale
AAE-29955

tom-dal:
AAE-29955


jesty (approved):
LGTM

PR #1696, opened 2025-02-13
AAE-31637 Spring boot update with hibernate downgrade

wojciech-piotrowiak:


activiti-cloud-service-common/activiti-cloud-services-swagger/src/main/java/org/activiti/cloud/common/swagger/springdoc/customizer/CustomOperationIdCustomizer.java
```diff
@@ -0,0 +1,28 @@
...
+
+package org.activiti.cloud.common.swagger.springdoc.customizer;
+
+import io.swagger.v3.oas.models.OpenAPI;
+import org.springdoc.core.customizers.OperationIdCustomizer;
+
+public class CustomOperationIdCustomizer extends OperationIdCustomizer {
+
+    @Override
+    public void customise(OpenAPI openAPI) {
```
Giovanni007, line 25: After upgrading versions, `NamingOperationCustomizer` was executed before the default  Springdoc default `OperationIdCustomizer` that is adding `_1` to `operationId`. For this reason only solution I found was override the default bean

activiti-cloud-runtime-bundle-service/activiti-cloud-starter-runtime-bundle/src/test/java/org/activiti/cloud/starter/tests/helper/ProcessInstanceRestTemplate.java
```diff
@@ -152,12 +153,12 @@ private ResponseEntity<CloudProcessInstance> startCreatedProcessCall(String base
         );
     }
 
-    private ResponseEntity<ActivitiErrorMessageImpl> startCreatedProcessCallFail(String baseURL) {
+    private ResponseEntity<EntryResponseContent<ActivitiErrorMessageImpl>> startCreatedProcessCallFail(String baseURL) {
```
Giovanni007, line 156: In the previous Spring version, for some reason, the response was not always detected as application/json during certain tests. As a result, the wrong converter was picked—one of the default ones that we had explicitly excluded [here](https://github.com/Activiti/activiti-cloud/blob/develop/activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/main/java/org/activiti/cloud/alfresco/config/AlfrescoWebAutoConfiguration.java#L88), and so the test were passing.
In the latest version of Spring, the response is correctly identified as application/json, which causes [AlfrescoJackson2HttpMessageConverter](https://github.com/Activiti/activiti-cloud/blob/develop/activiti-cloud-service-common/activiti-cloud-services-dbp-rest/src/main/java/org/activiti/cloud/alfresco/converter/json/AlfrescoJackson2HttpMessageConverter.java) to be executed (the only one we allow to manage application/json responses). I believe this is the correct behavior from a configuration standpoint. This is further supported by the fact that we return an EntityModel in the [ControllerAdvices](https://github.com/Activiti/activiti-cloud/blob/develop/activiti-cloud-service-common/activiti-cloud-service-error-handlers/src/main/java/org/activiti/cloud/services/error/CommonExceptionHandler.java), and some [tests](https://github.com/Activiti/activiti-cloud/blob/develop/activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-rest-impl/src/test/java/org/activiti/cloud/services/rest/controllers/ProcessDefinitionControllerImplIT.java#L361) already explicitly expect entry to be returned for error responses.

Giovanni007 (approved):

PR #1721, opened 2025-03-13
AAE-32830 Update dependabot config 

wojciech-piotrowiak:


Giovanni007 (commented):
Is there a specific reason we keep more than five PRs open and continue allowing Dependabot to create more, instead of focusing on merging the existing ones?

wojciech-piotrowiak:
> Is there a specific reason we keep more than five PRs open and continue allowing Dependabot to create more, instead of focusing on merging the existing ones?

Usually, it is a good approach, but some PRs are really hard to finish, there are many breaking changes when actually it is still safe to use the old version so there is nothing to hurry. In the meantime, we might work on changes that are low hanging fruits or are related to security (but we can't because new PRs are not generated)

Giovanni007 (approved):

PR #1749, opened 2025-03-24
AAE-33087 Set Root Process Instance Id In ProcessInstanceEntity

revatilimaye:
Issue Link : https://hyland.atlassian.net/browse/AAE-33087

### DIFF
--- a/activiti-cloud-runtime-bundle-service/activiti-cloud-starter-runtime-bundle/src/test/java/org/activiti/cloud/starter/tests/services/audit/AuditProducerIT.java
+++ b/activiti-cloud-runtime-bundle-service/activiti-cloud-starter-runtime-bundle/src/test/java/org/activiti/cloud/starter/tests/services/audit/AuditProducerIT.java
@@ -916,30 +916,17 @@ public class AuditProducerIT {
         String processInstanceId = processInstance.getBody().getId();
 
         // when
-        List<String> subprocessIds = runtimeService
+        List<org.activiti.engine.runtime.ProcessInstance> childInstances = runtimeService
             .createProcessInstanceQuery()
             .superProcessInstanceId(processInstanceId)
-            .list()
-            .stream()
-            .map(it -> it.getProcessInstanceId())
-            .collect(Collectors.toList());
-        // then
+            .list();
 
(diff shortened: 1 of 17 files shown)


erdemedeiros (changes requested):
Please, add test coverage:
- You can add tests for `ProcessCreatedEventHandler` that are currently missing (similar to what was done for variables)
- You can also create a new test (or update one of the existing tests in AuditProducerIT) to make sure we are emitting events where root process instance id is set for process instance.

activiti-cloud-runtime-bundle-service/activiti-cloud-starter-runtime-bundle/src/test/java/org/activiti/cloud/starter/tests/services/audit/AuditProducerIT.java
```diff
@@ -923,11 +924,23 @@ void testTwoSubProcesses() {
...
             .collect(Collectors.toList());
         // then
+
+        List<String> rootProcessInstanceIds = runtimeService
+            .createProcessInstanceQuery()
+            .superProcessInstanceId(processInstanceId)
+            .list()
+            .stream()
+            .map(Execution::getRootProcessInstanceId)
+            .toList();
```
erdemedeiros: Better to avoid retrieving the list of child process instances twice: you can extract a variable `childInstances` and reuse it. Also, there is no need of collecting the ids inside a variable if you're only using it inside the assertion. You can use `extractingMethod` from `AssertJ` to achieve the same thing with less code. IMO, asserting that it has two entries is not essential and it should be enough asserting that it contains only the expected root process instance id:
```java
     List<org.activiti.engine.runtime.ProcessInstance> childInstances = runtimeService
            .createProcessInstanceQuery()
            .superProcessInstanceId(processInstanceId)
            .list();
```
```java
    assertThat(childInstances)
            .extracting(Execution::getRootProcessInstanceId)
            .containsOnly(processInstanceId);
```

erdemedeiros (commented):
I've added a suggestion of improvement for the test assertions

erdemedeiros (approved):
LGTM

PR #1771, opened 2025-04-03
AAE-32880 Constant and static values are not fetched properly

rishikabiswas:
Description: PR to fix issue with fetching process Definition static and constant values for start event.
Issue Link: https://hyland.atlassian.net/browse/AAE-32880


activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-core/src/main/java/org/activiti/cloud/services/core/ProcessDefinitionValuesService.java
```diff
@@ -92,7 +93,7 @@ public Map<String, Object> getProcessModelConstantValuesForStartEvent(String id)
 
     private ExtensionsStartEventId getProcessExtensionsForStartEvent(String id, boolean formRequired) {
         BpmnModel bpmnModel = repositoryService.getBpmnModel(id);
-        Process process = bpmnModel.getMainProcess();
+        Process process = bpmnModel.getProcessById(getProcessId(id));
```
jsokolowskii: Could you add a fallback to `getMainProcess` if `getProcessId` returns null? This should not happen, but better safe than sorry.

jsokolowskii (changes requested):
Please check the comment 🙏

jsokolowskii (approved):
LGTM!

PR #1785, opened 2025-04-11
AAE-32831 Add Activiti restricted key provider to GraphQl query configuration

igdianov:
This PR adds support for GraphQl Query user restrictions based on the existing Query Rest restrictions predicates rules and logic. The Query rest implementation configures restricted keys provider to apply primary keys restrictions on entity queries for the authenticated user security attributes, i.e. roles and groups.

It also enables support for the authenticated user principal role based GraphQl Query fields visibility at the runtime. 

The security configurations provides sensible defaults and can be adjusted for different deployment scenarios. 

Fixes https://hyland.atlassian.net/browse/AAE-32831


andrea-ligios (approved):
LGTM!

PR #1826, opened 2025-06-09
AAE-35605 Add index on audit tables for performance improvement

eromano:
Add index on audit tables for performance improvement


jesty (approved):
LGTM

PR #1845, opened 2025-07-02
AAE-36025 Filter form process definitions

Giovanni007:
Needs: https://github.com/Activiti/Activiti/pull/5067


activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-core/src/main/java/org/activiti/cloud/services/core/BaseProcessDefinitionService.java
```diff
@@ -27,6 +27,8 @@ public abstract class BaseProcessDefinitionService {
 
     private final List<ProcessDefinitionDecorator> processDefinitionDecorators;
 
+    protected static final String EXCLUDED_CATEGORY = "#triggerableByForm";
```
wojciech-piotrowiak: I'm not sure if categories are used only in processes if not then would be good to mention it

Giovanni007: Renamed

wojciech-piotrowiak (approved):

PR #1854, opened 2025-07-09
AAE-35733 - Extend TaskControllerAdvice in order to use it when the module is imported

tom-dal:
Controller Advice works fine in activiti-cloud but if not imported it will not work in projects that have activiti-cloud as dependency


jesty (approved):
LGTM

PR #1871, opened 2025-07-28
fix/AAE-36971 Fixing version issues that causes conflict

Sourav-Bhattacharyya:
Fixing the issues with the existing PR that is aimed at upgrading batik version. 
https://hyland.atlassian.net/browse/AAE-36971


souvikpanda99 (changes requested):
Remove the word fix for the PR title and name the PR as bump batik version to 1.19 to properly reflect what this PR is doing

PR #1873, opened 2025-07-29
AAE-37007 Add exclude category parameter to process definitions endpoints

graric:


andrea-ligios (approved):

Giovanni007 (approved):

activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-core/src/main/java/org/activiti/cloud/services/core/BaseProcessDefinitionService.java
```diff
@@ -52,4 +67,18 @@ protected ExtendedCloudProcessDefinition decorate(
...
+        if (validateInput(excludedCategory)) {
+            LOGGER.debug("Excluding process definitions with category: {}", excludedCategory);
+
+            processDefinitionsPayloadBuilder.withProcessCategoryToExclude(excludedCategory);
+        }
+        return processDefinitionsPayloadBuilder.build();
+    }
+
+    protected boolean validateInput(String excludedCategory) {
+        return StringUtils.isNotEmpty(excludedCategory) && excludedCategory.matches("[a-zA-Z0-9_\\-#.]+");
```
Giovanni007, line 82: I'm not sure if it's okay to have this pattern hardcoded. Is this the same pattern used by the frontend? If so, we're likely fine.

Giovanni007 (approved):
Only one small comment. LGTM

activiti-cloud-runtime-bundle-service/activiti-cloud-services-runtime-bundle/activiti-cloud-services-core/src/main/java/org/activiti/cloud/services/core/BaseProcessDefinitionService.java
alep85: FE send category as `category:"#triggerableByForm"` or `category: ''`

Giovanni007: Now that the filter is generic, any category created in UI can be passed to the backend (even though, for now, we’re only using #triggerableByForm) So I meant that the regex should match what we allow in the FE when setting a category when creating a process

wiktord2000:
Please don't merge it yet, we need to clarify sth - at first glance recent revert didn't fix fully the reported regression.

acichon-h (approved):
Tested on image generated by PR: https://github.com/Alfresco/hxp-process-services/pull/3747

`rb/v1/process-definitions` filtering ✅ 
`rb/admin/v1/process-definitions` filtering ✅ 
no filtering ✅ 

QA approved ✅

PR #1885, opened 2025-08-05
AAE-35904 Event handler failure should mark task as failed

wojciech-piotrowiak:


.github/workflows/main.yml
```diff
@@ -142,7 +142,7 @@ jobs:
 
   acceptance-tests:
     runs-on: ubuntu-latest
-    if: ${{ needs.pre-checks.outputs.skip != 'true' }}
+    if: ${{ needs.pre-checks.outputs.skip != 'true' && github.event.pull_request.draft == false }}
```
wojciech-piotrowiak, line 145: those changes speed up development if you don't want to run acceptance tests in draft state, LMK if this is OK to merge as it is

PR #1942, opened 2025-09-22
AAE-34337 Add ability to republish discarded events in order to be consumed by another node instead of retrying on the same node

jesty:
When we are doing the upgrade of an application and we have a pod started with a new version and a pod started with an older version it could happen that the new pod produces an event with then new event version that it is consumed by a pod still running with the old version.
This will discard the event leaving the service task in a started status that will never be completed.

In order to solve this issue we added the retry parameter to the ConnectorBinding annotation. This new attribute will be used where requrired, for instance when running a ScriptTask.
Thanks to this new attribute we will republish the event multiple time in order to try to consume from a different node.


igdianov (approved):
LGTM

tom-dal (approved):
LGTM

PR #1948, opened 2025-09-25
Trigram Search idea

CaptainGlac1er:
I'm curious if this could spark some ideas around enabling FTS in the query service. 

This might not be a working example

I added type into the query so the trigram would only index on string variables to limit index size. Then you could enable it by passing type string in the query to enable the FTS search. The same idea could carry over to the process variables as well.


CaptainGlac1er:
Another idea is that a new variable type could be added `fts` so you can select what variable should be available for FTS.

CaptainGlac1er:
Another is that an additional search type can be added `FTS` like `LIKE` but then the `tsvector` approach could be used

CaptainGlac1er:
Another example using tsvector
https://github.com/Activiti/activiti-cloud/pull/1949

PR #1953, opened 2025-09-29
AAE-38546 Springdoc upgrade to 2.8.9

wojciech-piotrowiak:


activiti-cloud-service-common/activiti-cloud-services-swagger/src/main/java/org/activiti/cloud/common/swagger/springdoc/conf/SwaggerAutoConfiguration.java
```diff
@@ -62,6 +65,12 @@ public BaseOpenApiBuilder baseOpenApiBuilder(BuildProperties buildProperties, OA
         return new BaseOpenApiBuilder(buildProperties, swaggerOAuthFlow);
     }
 
+    @Bean(GLOBAL_OPEN_API_CUSTOMIZER)
```
wojciech-piotrowiak, line 68: this forces overload to CustomOperationIdCustomizer which stops modifications on operationID (default implementation is adding postfix like _1 if any operation id is duplicated and we want to avoid it)

activiti-cloud-service-common/activiti-cloud-services-swagger/src/test/resources/org/activiti/cloud/common/swagger/apidocs/springdoc-api-docs.json
```diff
@@ -61,6 +61,7 @@
             "name": "maxItems",
             "in": "query",
             "required": false,
+            "style": "form",
```
wojciech-piotrowiak, line 64: Those changes are generated, I'm not sure if we have to be backward compatible on each field

jsokolowskii (approved):

PR #1993, opened 2025-10-31
AAE-39613 Log error when invalid access token instance problem occurs

ffazzini:
The change would be consistent to existing validators like: `IsNotBeforeValidationCheck`
and being an error it would be log the problem at error level


jesty (approved):
LGTM

PR #2025, opened 2025-11-21
[AAE-40170] Update testcontainers version to 2.0.2

aonych:
Updating testcontainers version to fix issue with running test locally.
More info:
https://github.com/testcontainers/testcontainers-java/issues/11212


dsibilio:
@aonych FYI this DependaBot PR that does the same already exists: https://github.com/Activiti/activiti-cloud/pull/2011 unless we need to make more changes than expected here to make it work

PR #2041, opened 2025-12-02
AAE-39909 Bump com.github.dasniko:testcontainers-keycloak to 4.0.0

dsibilio:
Main changes are due to:
- removal of `keycloak-core` from test suite (`KeycloakPrincipal` does not exist anymore in test scope as the new `testcontainers-keycloak` does not bring it in transitively anymore)
- `testcontainers-keycloak` by default now assumes the new default management port `9000` should be used for the startup health check, this does not work well with [Keycloak < 25 as this does not expose management endpoints via a separate server.](https://www.keycloak.org/docs/latest/upgrading/index.html#management-port-for-metrics-and-health-endpoints)


aonych (approved):

mavotto (approved):

dsibilio:
**Note**: need to test it in the following places prior to merging https://github.com/search?q=org%3AAlfresco+%22activiti-cloud-services-test-containers%22&type=code&p=1

**UPDATE:** all tests passed.

PR #2098, opened 2026-01-09
AAE-41210 Remove unused Eureka module

killerboot:
The original idea wat to remove the whole metadata maven module, but the integration tests started to fail.
Instead only the [Eureka](https://github.com/Netflix/eureka) dependency was removed. It has old dependencies and is not very well maintained.


erdemedeiros (approved):
LGTM

dsibilio (approved):
LGTM

PR #2104, opened 2026-01-15
AAE-40609 Add skip-tests label support to skip tests on PR builds

souvikpanda99:
https://hyland.atlassian.net/browse/AAE-40609

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,11 +24,11 @@ env:
   GITHUB_PR_NUMBER: ${{ github.event.pull_request.number }}
 
 jobs:
-
   pre-checks:
     runs-on: ubuntu-latest
     outputs:
       skip: ${{ steps.dependabot.outputs.result }}
+      skip-tests: ${{ steps.check-skip-tests.outputs.result }}
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -105,7 +110,13 @@ jobs:
 
       - name: Maven Build and Test
         shell: bash
-        run: mvn -DskipAcceptanceTests=true -DunitTests.parallel=true -T 1C install ${{ env.MAVEN_CLI_OPTS }}
+        run: |
+          TEST_OPTS=""
+          if [[ "${{ needs.pre-checks.outputs.skip-tests }}" == "true" ]]; then
+            TEST_OPTS="-DskipTests=true"
```
atchertchian, line 116: this is a bit different on other builds we do, see https://github.com/Alfresco/alfresco-build-tools/blob/master/.github/actions/maven-configure/action.yml#L79 although I'm not clear here how integration tests are handled, I think they're in next step anyway

erdemedeiros: Here only adding `-DskipTests=true` seems to be enough for skipping all test tests

erdemedeiros (approved):
LGTM

PR #2196, opened 2026-03-17
AAE-43226 Bump Spring Boot to 4.0.4

shsahahyland:


activiti-cloud-build/activiti-cloud-build-dependencies-parent/pom.xml
```diff
@@ -121,12 +106,17 @@
       </dependency>
       <!-- AAE-42877 Update jackson to fix security issue -->
       <dependency>
-        <groupId>com.fasterxml.jackson</groupId>
+        <groupId>tools.jackson</groupId>
         <artifactId>jackson-bom</artifactId>
         <version>${jackson.version}</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
```
dsibilio: @shsahahyland FYI this was introduced in the first place to solve sec issues. If we're bumping Jackson to a vuln-free version then IMO we should get rid of this unnecessary dep management section entirely (repeated across several BE repos) cc: @killerboot @souvikpanda99

pom.xml
```diff
@@ -227,7 +227,7 @@
     <org-json.version>20250107</org-json.version>
 
     <!-- dependencies versions -->
-    <hibernate.version>7.2.7.Final</hibernate.version>
+    <hibernate.version>7.2.9.Final</hibernate.version>
     <!-- Temporary Dependency. Remove when Hibernate version for Java 25 is available -->
```
graric, line 231: Should this dependency and this comment be removed after the upgrade?

igdianov: It should be pinned until the next Spring Boot 4 version updates its dependency BoM versions with new Hibernate version.

andrea-ligios (approved):

PR #2213, opened 2026-03-26
AAE-43788 Add dependency graph submission

killerboot:


.github/workflows/dependency-graph-submission.yml
```diff
@@ -0,0 +1,31 @@
...
+  push:
+    branches: ["develop"]
+jobs:
+  submit:
+    name: Submit
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: actions/cache@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5.0.4
```
atchertchian: I wonder if you want only to restore (and not save cache). given that key will probably match the main maven build

killerboot: TBH I copied this from https://github.com/Activiti/Activiti/blob/develop/.github/workflows/dependency-graph-submission.yml 
Also a similar reference from the main workflow in this repo.
Should it be done differently?

atchertchian: The problem is not about not restoring the cache, it's about saving it with same key than main regular build: do we want to have a specific cache for both or do we want to share the same and potentially step on each other?
Two options:
1. have a specific cache -> make the key specific
2. reuse the main build cache -> make this action only do the restore
Activiti has the same issue.

Note that on other repos (through shared gh actions). we do the cache operation in two steps (restore and save) so that save is not done on PRs.

dsibilio (approved):
LGTM

.github/workflows/dependency-graph-submission.yml
```diff
@@ -0,0 +1,31 @@
+name: "Dependency Graph Submission of Transitive Dependencies"
+
+on:
+  push:
+    branches: ["develop"]
```
atchertchian: let's test this here
```suggestion
    branches: ["develop", "feature/AAE-43788-add-dependency-graph"]
```

.github/workflows/dependency-graph-submission.yml
```diff
@@ -0,0 +1,31 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: actions/cache/restore@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5.0.4
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
```
atchertchian:
```suggestion
```

(because actually maven-dependency-scan calls maven-configure which already handles this restore)

atchertchian (approved):

PR #2217, opened 2026-03-31
AAE-43932 Fix shared variables between tests

killerboot:


killerboot:
The suggested approach did not work, transferring the work to https://github.com/Activiti/activiti-cloud/pull/2218

PR #2252, opened 2026-04-25
AAE-43226 Preserve null map entries in payload variables only

astrologic7:


andrea-ligios (approved):

erdemedeiros (approved):
Let's merge as it is for unblock the pipeline, but let's re-work the code just after to avoid optional dependency

PR #2262, opened 2026-04-30
AAE-45142 Handle fallback for connectors not setting content type

erdemedeiros:
Spring Cloud Function 5 / Stream 5 (Spring Boot 4) no longer re-applies the binding-level contentType on the router-mediated invocation path; the inbound `MessageHeaders.CONTENT_TYPE` is now authoritative. Messages arriving via RabbitMQ
  with `contentType=application/octet-stream` bypass JsonMessageConverter, get claimed by `ByteArrayMessageConverter`, and surface to `Consumer<Message<T>>` beans as raw `byte[]`, triggering a `ClassCastException`.

  Introduce `MessageContentTypeNormalizer` and apply it on both the router (`FunctionRouterConfiguration`) and direct binding (`FunctionBindingConfiguration`) paths. When the inbound contentType header is missing or is not compatible with the expected content type, it  is overridden with expected content type (defaulted to `application/json` if missing), before the function is invoked; other content types are preserved.


igdianov (approved):
LGTM

graric (approved):

PR #2342, opened 2026-06-03
AAE-46649 Fix flaky runtime acceptance tests

annkarwatka:
Problem: Several Playwright tests intermittently failed with Unable to find process instance for the given id:'…' or undefined task-field reads. Root cause: tests called QueryService.getProcessInstance(...) / getTaskById(...) immediately after a runtime-bundle write, before the messaging pipeline (RabbitMQ/Kafka) had propagated the event into the query-service projection.

Fix applied: Wrap the affected reads in expect.poll(..., pollOptions('querySync')) callbacks that tolerate the transient "not found" / undefined state, matching the proven pattern already used in signal-actions / timer-actions / service-tasks-actions.


MichalFidor (approved):
For now LGTM. For the future, we will need to extract the logic from pools and move it to services. But this is a generic thing to accomplish, so approving it right now.

PR #2355, opened 2026-06-10
AAE-46511 Improve guardrail audit error message

mikolajbrzezinski:
Adding a new Integration Error constructor so that we can surface custom message instead of the root error message in audit

### DIFF
--- a/activiti-cloud-api/activiti-cloud-api-process-model-impl/src/test/java/org/activiti/cloud/api/process/model/impl/IntegrationErrorImplTest.java
+++ b/activiti-cloud-api/activiti-cloud-api-process-model-impl/src/test/java/org/activiti/cloud/api/process/model/impl/IntegrationErrorImplTest.java
@@ -272,22 +272,14 @@ class IntegrationErrorImplTest {
         assertThat(result.getErrorMessage()).isEqualTo(customMessage);
     }
 
-    @Test
-    void should_fallbackToDetailedErrorMessage_when_customMessageIsNull() {
+    @ParameterizedTest
+    @NullSource
+    @ValueSource(strings = { "", "   " })
+    void should_fallbackToDetailedErrorMessage_when_customMessageIsNullOrBlank(String customMessage) {
         var rootCause = new RuntimeException("Root cause message");
         var error = new RuntimeException("Error message", rootCause);
 
-        var result = new IntegrationErrorImpl(integrationRequest, error, null);
-
-        assertThat(result.getErrorMessage()).isEqualTo("Error message caused by: Root cause message");
-    }
-
-    @Test
-    void should_fallbackToDetailedErrorMessage_when_customMessageIsBlank() {
-        var rootCause = new RuntimeException("Root cause message");
-        var error = new RuntimeException("Error message", rootCause);
-
-        var result = new IntegrationErrorImpl(integrationRequest, error, "   ");
+        var result = new IntegrationErrorImpl(integrationRequest, error, customMessage);
 
         assertThat(result.getErrorMessage()).isEqualTo("Error message caused by: Root cause message");
     }
--- a/activiti-cloud-api/activiti-cloud-api-process-model-impl/src/test/java/org/activiti/cloud/api/process/model/impl/IntegrationErrorImplTest.java
+++ b/activiti-cloud-api/activiti-cloud-api-process-model-impl/src/test/java/org/activiti/cloud/api/process/model/impl/IntegrationErrorImplTest.java
@@ -260,4 +260,35 @@ class IntegrationErrorImplTest {
 
         assertThat(result.getErrorMessage()).isEqualTo("actual error");
     }
+
+    @Test
+    void should_returnCustomMessage_when_customMessageIsProvided() {
+        var rootCause = new RuntimeException("Root cause message");
+        var error = new RuntimeException("Error message", rootCause);
+        String customMessage = "Request content violates guardrail policy";
+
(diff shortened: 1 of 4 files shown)


activiti-cloud-connectors/activiti-cloud-starter-connector/src/main/java/org/activiti/cloud/connectors/starter/model/IntegrationErrorBuilder.java
```diff
@@ -52,7 +53,17 @@ public IntegrationError build() {
...
+        IntegrationErrorImpl integrationError;
+        if (
+            error instanceof CloudBpmnError cloudBpmnError &&
+            "BLOCKED_BY_GUARDRAIL".equals(cloudBpmnError.getErrorCode()) &&
+            cloudBpmnError.getMessage() != null
+        ) {
+            integrationError = new IntegrationErrorImpl(integrationRequest, error, cloudBpmnError.getMessage());
+        } else {
+            integrationError = new IntegrationErrorImpl(integrationRequest, error);
+        }
```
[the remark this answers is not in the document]
mikolajbrzezinski: this is the whole point of the PR

activiti-cloud-api/activiti-cloud-api-process-model-impl/src/test/java/org/activiti/cloud/api/process/model/impl/IntegrationErrorImplTest.java
```diff
@@ -260,4 +260,35 @@ void should_returnErrorMessage_when_fqcnStrippingLeavesBlankRootMessage() {
...
+        var rootCause = new RuntimeException("Root cause message");
+        var error = new RuntimeException("Error message", rootCause);
+
+        var result = new IntegrationErrorImpl(integrationRequest, error, null);
+
+        assertThat(result.getErrorMessage()).isEqualTo("Error message caused by: Root cause message");
+    }
+
+    @Test
+    void should_fallbackToDetailedErrorMessage_when_customMessageIsBlank() {
```
aonych: We can remove this test and make one parameterizedTest for both cases by adding:
`@ParameterizedTest
@NullSource
@ValueSource(strings = { "", "   " })`

aonych (changes requested):

activiti-cloud-api/activiti-cloud-api-process-model-impl/src/test/java/org/activiti/cloud/api/process/model/impl/IntegrationErrorImplTest.java
mikolajbrzezinski: Smart! Thank you, will do

aonych (approved):

activiti-cloud-connectors/activiti-cloud-starter-connector/src/main/java/org/activiti/cloud/connectors/starter/model/IntegrationErrorBuilder.java
```diff
@@ -52,7 +55,15 @@ public IntegrationError build() {
         Objects.requireNonNull(integrationRequest);
         Objects.requireNonNull(error);
 
-        IntegrationErrorImpl integrationError = new IntegrationErrorImpl(integrationRequest, error);
+        IntegrationErrorImpl integrationError;
+        if (
+            error instanceof CloudBpmnError cloudBpmnError && BLOCKED_BY_GUARDRAIL.equals(cloudBpmnError.getErrorCode())
+        ) {
+            integrationError = new IntegrationErrorImpl(integrationRequest, error, cloudBpmnError.getMessage());
```
dsibilio: Why are we doing this logic here? We're leaking details about our connectors and their error codes all the way down in `activiti-cloud -> IntegrationErrorBuilder` + adding 1 more repo where we have to sync error codes across projects.

Shouldn't we just customize the builder so that it can optionally allow setting a custom error message (default to behaving as it did before otherwise) and set it in the connector code itself to build/reconstruct the exception using the +1param constructor when we need to surface a custom error message?

See how other Connectors normally operate: github.com/search?q=repo%3AAlfresco%2Fhxp-process-services%20integrationerrorbuilder&type=code

dsibilio (dismissed):
Approach doesn't look good to me, please give a look 🙏

activiti-cloud-connectors/activiti-cloud-starter-connector/src/main/java/org/activiti/cloud/connectors/starter/model/IntegrationErrorBuilder.java
```diff
@@ -48,11 +49,23 @@ public static IntegrationErrorBuilder errorFor(
         return new IntegrationErrorBuilder(integrationRequest, connectorProperties, error);
     }
 
+    public static IntegrationErrorBuilder errorFor(
```
aonych: Instead of creating this method, we can add a setter on the existing builder, like:
`public IntegrationErrorBuilder withCustomErrorMessage(String customErrorMessage) {
    this.customErrorMessage = customErrorMessage;
    return this;
}`
We can than use it as:
`IntegrationErrorBuilder
    .errorFor(integrationRequest, connectorProperties, e)
    .withCustomErrorMessage(e.getMessage())
    .buildMessage();`

aonych (changes requested):

aonych (approved):
LGTM! Thanks :)

PR #2358, opened 2026-06-10
AAE-46521 Persist process variables history

jsokolowskii:
This PR introduces persistence of **process variable change history** in the Query Service, so create/update/delete variable events result in an append-only history record that can be queried later.

**Changes:**
- Add `process_variable_history` persistence model + repository and Liquibase schema updates.
- Extend process-variable create/update/delete event handling to persist history entries (skipping ephemeral variables).
- Add/expand unit + integration tests to verify history persistence behavior and routing.
- The history persistence is disabled by default and can be enabled via `query.process-variable-history` feature toggle


elena-grosu-h (approved):
LGTM

PR #2370, opened 2026-06-12
AAE-41740 Publish query events after query DB commit

sdonnell-hyland:
**Issue Link**
https://hyland.atlassian.net/browse/AAE-41740

This PR updates the Query consumer to forward the original engine event batch as `queryEvents` only after the query-side transactional work has completed.

It also includes:
- a dedicated `queryEventsProducer` binding and downstream configuration updates to consume `queryEvents`
- queued forwarding support for starter-based deployments using a persistent channel message store
- Liquibase/H2 schema updates for the internal query events message store
- dedicated integration and message-handler test coverage, including `QueryEventsProducerIT`


igdianov (approved):
LGTM

PR #2481, opened 2026-07-21
AAE-48833 Ensure runtime producer bindings use RabbitMq transacted message publisher

igdianov:
This PR updates runtime-bundle messaging configuration to ensure connector producer bindings are configured for RabbitMQ transacted publishing, and aligns destination resolution with the function-router binding model.

**Changes:**
- Sets `spring.cloud.stream.rabbit.bindings.<binding>.producer.transacted=true` for several producer bindings across runtime-related starter/property modules.
routing is enabled.
- Sets `auditProducerIncidents` explicitly non-transacted if somebody decided to set defaults to transacted=true 
- Add an auto-configuration initializer that marks Rabbit connector producer bindings as `transacted=true` for all discovered connector implementations.
- Add transacted configuration properties that marks Rabbit connector producer bindings as `transacted=true` for all discovered connector implementations.
- Update example/runtime tests and dependencies to include and validate Rabbit binder extended producer properties (including transacted connector bindings).
- Switch message-event destination resolution from `BindingServiceProperties` to `FunctionBindingConfiguration.BindingResolver` to correctly resolve destinations when function 

https://hyland.atlassian.net/browse/AAE-48833

### DIFF
--- a/activiti-cloud-examples/example-runtime-bundle/starter/src/test/resources/logback-test.xml
+++ b/activiti-cloud-examples/example-runtime-bundle/starter/src/test/resources/logback-test.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration>
   <include resource="org/springframework/boot/logging/logback/base.xml" />
-  <root level="OFF" />
+  <root level="INFO" />
 </configuration>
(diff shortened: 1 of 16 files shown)


activiti-cloud-examples/example-runtime-bundle/starter/src/test/resources/logback-test.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration>
   <include resource="org/springframework/boot/logging/logback/base.xml" />
-  <root level="OFF" />
```
graric, line 4: Do we wanna keep them enabled?

graric (approved):
Great work!

erdemedeiros (approved):

PR #2497, opened 2026-07-22
AAE-47841 Increase timeout to minimize parallel process errors

mikolajbrzezinski:
Issue link - https://hyland.atlassian.net/browse/AAE-47841
This solves the legitimate timeout in case of too much load, but the self-deadlock case needs much more work to resolve.


igdianov (approved):
LGTM

dsibilio (approved):
🍀

PR #2499, opened 2026-07-22
AAE-42058 Process stuck with application upgrade

jsokolowskii:
**Root cause**

When the Function Router is enabled, each connector message is dispatched to a per-connector, single-thread executor in a fire-and-forget way: the AMQP listener hands the message to the executor and returns immediately. With the previous auto-ack behaviour, RabbitMQ acknowledged the message the moment that handoff returned — before the work had actually run. During a rolling upgrade the old pod's executor is saturated or shutting down, so the handoff fails, but the message had already been acked and was therefore lost. The waiting service task never received a result, leaving the process instance stuck. In some cases the failure also surfaced as a false application error on the service task, or as an IllegalStateException: Channel closed; cannot ack/nack.

**Fix — overview**

Switch the Function Router's consumer to manual acknowledgement and only ack once the work genuinely completes; nack-and-requeue genuine delivery failures so the broker redelivers them instead of dropping them. All changes are in the shared activiti-cloud-service-messaging-config module (plus one runtime-bundle test).

**Changes (why / how)**

Manual ack for the Function Router consumer — FunctionRouterConfiguration

New functionRouterListenerContainerCustomizer bean sets AcknowledgeMode.MANUAL on the container of the function-router consumer group. It lives in FunctionRouterConfiguration (which is @ConditionalOnProperty(function-router.enabled)), so manual ack is only ever applied when the router — the component that performs the ack/nack — is actually present. When the router is disabled, nothing changes (avoids leaving a consumer on manual ack with nobody to acknowledge it).

Acknowledge on real outcome, nack+requeue on delivery failure — FunctionRouterConfiguration

After the async processing chain resolves: ack on genuine success or a genuine execution error (already delivered to the connector's error handler); nack with requeue=true on a delivery failure, so RabbitMQ redelivers the message rather than the router optimistically acking work it never finished. Ack/nack are applied to the original delivered message via its amqp_channel/amqp_deliveryTag headers; a no-op when those aren't present, so it's safe under any binder/ack configuration.

Delivery-failure detection & filtering — FunctionRouterConfiguration

RejectedExecutionException (queue full) and ImmediateRequeueAmqpException (executor shutting down) are recognised as delivery failures and are re-thrown / filtered out of the error path, so they are requeued by AMQP rather than being reported to the service task as application errors.

submitFunctionRequest wraps CompletableFuture.supplyAsync(...) in try/catch: the executor's rejection handler throws synchronously on the calling thread, so without the wrap the exception would bypass the retry and delivery-failure handling. It's converted to a failed future so it flows through the normal pipeline.

Executor lifecycle & rejection semantics — FunctionRouterExecutorFactory

Implements SmartLifecycle and shuts its executors down in a late phase, so in-flight work drains during shutdown. On rejection due to shutdown/interruption it now throws ImmediateRequeueAmqpException (a requeue signal) instead of a plain RejectedExecutionException.

Retry self-deadlock fix — CompletableFutureRetry

The retry helper re-submitted the retry inline on whatever thread completed the previous attempt — which is the executor's own single worker thread — causing it to block on its own full bounded queue (queue.offer(requestTimeout)), a self-deadlock (~5 min hang). Retries are now scheduled on the retry scheduler thread, and the configured back-off delay is actually applied.

Do not leak transport headers into the routed message — FunctionRouterConfiguration

Manual ack causes AmqpInboundChannelAdapter to put the live (non-serializable) AMQP Channel in the amqp_channel header. Consumers that persist the message (e.g. the messages-service aggregator backed by JdbcMessageStore, which serializes the message) failed on that header. The router now strips amqp_channel/amqp_deliveryTag from the message it forwards to the connector, while keeping them on the original message for ack/nack.

Connector retry off-by-one — ConnectorConfiguration

Fixed the retry-limit boundary (retryCount < maxRetry, previously maxRetry - 1) and only builds the retry message when a retry will actually happen; corrected a log-message typo.

**Tests**

FunctionRouterRabbitDeliveryFailureIT (new) — real-broker (RabbitMQ Testcontainers) coverage: queue-overflow is absorbed and redelivered; a permanently-shut-down executor's message is genuinely redelivered and succeeds once the executor recovers; delivery failures never leak to the error handler or the global error channel; a slow connector doesn't block unrelated connectors.

FunctionRouterDeliveryFailureIT (new) — unit-level coverage of the synchronous-rejection / requeue semantics.

Updated FunctionRouterBindingConfigurationIT, FunctionRouterExecutorFactoryTest, ConnectorConfigurationIT for the new behaviour, and OrphanedIntegrationRecoveryIT (both function-router-enabled and -disabled paths now defer ack until processing completes, so the broker redelivers on crash either way).

Verified passing: messaging-config suite (unit + ITs incl. the real-broker IT), MessageEventsFunctionRouterIT (17/17), RuntimeBundleFunctionRouterEnabledIT (15/15), OrphanedIntegrationRecoveryIT (3/3).

**Known follow-up (not addressed here)**

A delivery failure that never clears produces a tight nack → redelivery → re-rejection loop with no backoff for as long as the condition persists (documented in the IT). Worth a separate ticket for backoff/dead-lettering.


activiti-cloud-service-common/activiti-cloud-service-messaging-config/src/main/java/org/activiti/cloud/common/messaging/config/FunctionRouterExecutorFactory.java
```diff
@@ -56,7 +61,7 @@ public FunctionRouterExecutorFactory(Duration timeout) {
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
 
-            throw new RejectedExecutionException("Interrupted while waiting for queue capacity", e);
+            throw new ImmediateRequeueAmqpException("Interrupted during shutdown; requeueing for redelivery", e);
```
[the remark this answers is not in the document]
jsokolowskii: Fixed

activiti-cloud-service-common/activiti-cloud-service-messaging-config/src/main/java/org/activiti/cloud/common/messaging/config/FunctionRouterConfiguration.java
```diff
@@ -135,9 +138,7 @@ Consumer<Message<?>> functionRouterAnonymousConsumer(BiConsumer<Message<?>, Stri
 
     @Bean
     @ConditionalOnMissingBean
-    Function<String, ExecutorService> functionRouterExecutorFactory(
-        ActivitiCloudMessagingProperties messagingProperties
-    ) {
+    FunctionRouterExecutorFactory functionRouterExecutorFactory(ActivitiCloudMessagingProperties messagingProperties) {
```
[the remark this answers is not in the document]
jsokolowskii: Fixed

activiti-cloud-service-common/activiti-cloud-service-messaging-config/src/main/java/org/activiti/cloud/common/messaging/config/CompletableFutureRetry.java
```diff
@@ -55,9 +56,27 @@ private static <T> CompletableFuture<T> supplyAsyncWithRetry(
...
                 if (currentAttempt < maxRetries) {
                     log.debug("Attempt {} of {} failed. Retrying in {}", currentAttempt + 1, maxRetries, delay);
-                    return java.util.concurrent.CompletableFuture.supplyAsync(() -> null, scheduler).thenCompose(
-                        ignored -> supplyAsyncWithRetry(supplier, maxRetries, delay, currentAttempt + 1)
+                    // Schedule the retry on the scheduler thread (after the delay) instead of
+                    // composing it inline. The previous attempt completes on the per-registration
+                    // executor's own worker thread, and re-invoking the supplier there would
+                    // resubmit to that same single-thread executor - whose backpressure handler
+                    // blocks waiting for the queue to drain, deadlocking the worker against itself.
+                    var retry = new CompletableFuture<T>();
```
igdianov, line 64: Good catch!

activiti-cloud-service-common/activiti-cloud-service-messaging-config/src/main/java/org/activiti/cloud/common/messaging/config/FunctionRouterConfiguration.java
```diff
@@ -121,6 +128,28 @@ DeclarableCustomizer functionRouterAnonymousQueueCustomizer(ActivitiCloudMessagi
...
     }
 
+    /**
+     * Switches the function router's consumer to manual ack so {@code functionRouterMessageHandler}
+     * can ack/nack once the fire-and-forget handoff to the per-connector executor actually
+     * completes. Only registered when the function router is enabled (the enclosing configuration is
+     * conditional on that), so no consumer is ever left on manual ack with nothing to acknowledge it.
+     */
+    @Bean
+    ListenerContainerCustomizer<MessageListenerContainer> functionRouterListenerContainerCustomizer(
```
igdianov: This couples RabbitMq message listener with the Function Router implementation, which also need to support Kafka message listeners with the same functionality.

activiti-cloud-service-common/activiti-cloud-service-messaging-config/src/main/java/org/activiti/cloud/common/messaging/config/FunctionRouterConfiguration.java
```diff
@@ -204,19 +233,39 @@ BiConsumer<Message<?>, String> functionRouterMessageHandler(
...
                             )
                                 .setHeader(FunctionProperties.FUNCTION_DEFINITION, functionRegistration)
+                                // Manual ack (see the listener container customizer) puts the live,
+                                // non-serializable AMQP channel in the message headers so this router
+                                // can ack/nack the original message. Those transport headers must not
+                                // leak into the routed business message: a handler that persists it
+                                // (e.g. the messages-service's JdbcMessageStore aggregator serializes
+                                // the message) would fail to serialize the channel. The outer message
+                                // retains them for acknowledge()/negativelyAcknowledgeAndRequeue().
+                                .removeHeader(AmqpHeaders.CHANNEL)
```
igdianov: Hard coupling with RabbitMq

activiti-cloud-service-common/activiti-cloud-service-messaging-config/src/main/java/org/activiti/cloud/common/messaging/config/FunctionRouterConfiguration.java
```diff
@@ -230,6 +279,20 @@ BiConsumer<Message<?>, String> functionRouterMessageHandler(
...
                                     .exceptionally(error -> {
+                                        var cause = error instanceof CompletionException ce ? ce.getCause() : error;
+
+                                        // Delivery failures must be requeued by AMQP, not sent to
+                                        // the service task as errors. Rethrow so the outer handler
+                                        // nacks. debug, not warn: an unclearing failure loops with
+                                        // no backoff and would flood the logs.
+                                        if (
+                                            cause instanceof RejectedExecutionException ||
+                                            cause instanceof ImmediateRequeueAmqpException
```
igdianov: It is better to handle any AMQP specific exceptions within a binder message listener container, to keep the router binder agnostic

activiti-cloud-service-common/activiti-cloud-service-messaging-config/src/main/java/org/activiti/cloud/common/messaging/config/FunctionRouterConfiguration.java
```diff
@@ -302,11 +388,51 @@ BiConsumer<Message<?>, String> functionRouterMessageHandler(
...
+                log.warn("Failed to acknowledge message {}", message, e);
+            }
+        });
+    }
+
+    /**
+     * Nacks the message with {@code requeue=true} when delivered with a manual-ack
+     * channel/delivery-tag so AMQP redelivers it; a no-op otherwise.
+     */
+    private static void negativelyAcknowledgeAndRequeue(Message<?> message) {
```
igdianov: It will be problematic to requeue any messages, because, requeued messages will be replayed again for all subscribers inside function router. All consumers need to be idempotent to be handle requeued messages

activiti-cloud-service-common/activiti-cloud-service-messaging-config/src/main/java/org/activiti/cloud/common/messaging/config/FunctionRouterConfiguration.java
```diff
@@ -302,11 +388,51 @@ BiConsumer<Message<?>, String> functionRouterMessageHandler(
...
                     }
                 );
         };
     }
 
+    /**
+     * Acks the message when delivered with a manual-ack channel/delivery-tag; a no-op otherwise, so
+     * it is safe to call regardless of ack-mode.
+     */
+    private static void acknowledge(Message<?> message) {
```
igdianov: Hard coupling with RabbitMq messaging. It should be applied using binder specific error handling framework

igdianov (changes requested):
The deadlock resolution looks good, but the function router should be decoupled from Rabbit specific message ack/nack to be able to support Kafka as well.

activiti-cloud-service-common/activiti-cloud-service-messaging-config/src/main/java/org/activiti/cloud/common/messaging/config/FunctionRouterConfiguration.java
jsokolowskii: Hard coupling has been resolved by injecting a DeliveryAcknowledgment bean.

activiti-cloud-service-common/activiti-cloud-service-messaging-config/src/main/java/org/activiti/cloud/common/messaging/config/FunctionRouterConfiguration.java
jsokolowskii: Hard coupling has been resolved by injecting a DeliveryAcknowledgment bean.

activiti-cloud-service-common/activiti-cloud-service-messaging-config/src/main/java/org/activiti/cloud/common/messaging/config/FunctionRouterConfiguration.java
jsokolowskii: The exception was changed to a binder-agnostic one.

activiti-cloud-service-common/activiti-cloud-service-messaging-config/src/main/java/org/activiti/cloud/common/messaging/config/FunctionRouterConfiguration.java
jsokolowskii: The requeue has been changed in a way that only takes into account registrations that should reprocess the message pinned via
 the TARGET_REGISTRATIONS header.

activiti-cloud-service-common/activiti-cloud-service-messaging-config/src/main/java/org/activiti/cloud/common/messaging/config/FunctionRouterConfiguration.java
jsokolowskii: Hard coupling has been resolved by injecting a DeliveryAcknowledgment bean.

PR #2571, opened 2026-08-26
AAE-50657 Bump amqp-client to 5.33.0

DavidOlson-Hyland:
This is to fix six CVEs (three HIGH, two MEDIUM, one LOW) all against com.rabbitmq:amqp-client.

https://hyland.atlassian.net/browse/AAE-50657


DavidOlson-Hyland:
Need to fix typo in branch name

PR #2577, opened 2026-09-01
AAE-47554 Add actionlint to pre-commit config

andrphilippi:
## Summary

Adds `actionlint` to the pre-commit config so GitHub Actions workflows are linted alongside the existing hooks, and fixes everything it surfaced.

- Adds the `rhysd/actionlint` hook, pinned to a commit SHA with a `# frozen: v1.7.12` comment, excluding the auto-generated `*.lock.yml` files.
- Fixes the shellcheck findings in `main.yml` and `create-github-prerelease.yaml`: quoting `$GITHUB_OUTPUT`, `$TAG`, `$VERSION`, `$IS_PREVIEW`, `$GITHUB_EVENT_NAME` and the branch-name variables (SC2086), plus the `git branch --show-current` substitution (SC2046).
- Reworks the `Export build version` step to pass `github.event_name`, the tag version, the PR labels and the PR number through `env:` instead of interpolating expressions straight into the script. actionlint masks each expression with a fixed-length placeholder before running shellcheck, which made the `preview` label comparison look statically unsatisfiable (SC2193). Going through `env:` removes the warning at its root rather than suppressing it, and also drops the expression-injection surface from PR label names. Behaviour is unchanged.
- Adds `.github/actionlint.yaml` to ignore `unexpected key "queue" for "concurrency" section`. `concurrency.queue` is [valid GitHub Actions syntax](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax) that actionlint v1.7.12 does not recognise yet. It is used intentionally in `_reusable-playwright-tests.yml` (AAE-48234) and is also emitted by the gh-aw compiler in `supply-chain-review.lock.yml`. There is no upstream issue for it and the actionlint author is currently inactive, so an ignore is the only available route.

Related Jira task: [AAE-47554](https://hyland.atlassian.net/browse/AAE-47554)

## Test plan

- [x] `pre-commit run actionlint --all-files` passes.
- [x] Verified with a real `shellcheck` v0.11.0 binary on `PATH` so the `SC*` rules actually ran. actionlint skips them silently when shellcheck is absent, which is what a bare Windows checkout would otherwise do; GitHub's Ubuntu runners have it preinstalled, so this matches what CI enforces.
- [x] `pre-commit run --files` across all touched files passes for every other hook too.
- [ ] CI pre-commit job is green.

[AAE-47554]: https://hyland.atlassian.net/browse/AAE-47554?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.pre-commit-config.yaml
```diff
@@ -40,3 +40,8 @@ repos:
           - java
           - xml
         args: ["--config", ".semgrep.yaml", "--error", "--metrics=off"]
+  - repo: https://github.com/rhysd/actionlint
+    rev: 914e7df21a07ef503a81201c76d2b11c789d3fca # frozen: v1.7.12
+    hooks:
+      - id: actionlint
+        exclude: '\.lock\.yml$'
```
andrphilippi:
```suggestion
        exclude: '^\.github/workflows/.*\.lock\.yml$'
```

.pre-commit-config.yaml
```diff
@@ -26,17 +26,22 @@ repos:
           - prettier-plugin-java@2.8.1
         exclude: ^\.github/workflows/.*\.(lock\.yml|md)$
   - repo: https://github.com/sirosen/check-jsonschema
-    rev: 0.37.4
+    rev: 12e63946db2c5cfcc9030fac21c3883ba88c6ed8 # frozen: 0.38.0
```
andrphilippi, line 29: Upgrade from 0.37.4 -> 0.38.0 to align to other web repos.

.pre-commit-config.yaml
```diff
@@ -26,17 +26,22 @@ repos:
...
       - id: semgrep
         types_or:
           - java
           - xml
         args: ["--config", ".semgrep.yaml", "--error", "--metrics=off"]
+  - repo: https://github.com/rhysd/actionlint
+    rev: 914e7df21a07ef503a81201c76d2b11c789d3fca # frozen: v1.7.12
+    hooks:
+      - id: actionlint
+        exclude: '^\.github/workflows/.*\.lock\.yml$'
```
atchertchian, line 47: Isn't this exclude making .actionlint.yaml unnecessary?

JWollweber: The .actionlint.yaml configures an exception for the queue keyword which is not known to actionlint yet, right? So I think both settings are needed.

JWollweber (approved):

.pre-commit-config.yaml
andrphilippi: Yes like Jonas said the one (.actionlint.yaml) is for suppressing specifically the other (exclude) is for ignoring files by pattern completely

atchertchian: I was asking in case actionlint exclude was needed for this lock file only :+1:

.github/dependabot.yml
```diff
@@ -67,6 +67,19 @@ updates:
...
+      interval: "weekly"
+    cooldown:
+      default-days: 3
+    groups:
+      pre-commit:
+        patterns:
+          - "*"
+        update-types:
+          - "minor"
+          - "patch"
```
atchertchian:
```suggestion
```

(is already present just below)

PR #2592, opened 2026-09-08
AAE-50838 Update spring-cloud.version to 2025.1.3

astrologic7:
https://hyland.atlassian.net/browse/AAE-50838

### DIFF
--- a/activiti-cloud-runtime-bundle-service/activiti-cloud-starter-runtime-bundle/src/test/resources/application-test.properties
+++ b/activiti-cloud-runtime-bundle-service/activiti-cloud-starter-runtime-bundle/src/test/resources/application-test.properties
@@ -100,3 +100,4 @@ spring.cloud.stream.bindings.auditConsumer.consumer.partitioned=${activiti.cloud
 #logging.level.org.activiti.engine.impl.db=trace
 #logging.level.org.activiti.engine.integration=trace
 #logging.level.org.springframework.retry=debug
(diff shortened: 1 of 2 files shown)


activiti-cloud-runtime-bundle-service/activiti-cloud-starter-runtime-bundle/src/test/resources/application-test.properties
```diff
@@ -100,3 +100,4 @@ spring.cloud.stream.bindings.auditConsumer.consumer.partitioned=${activiti.cloud
 #logging.level.org.activiti.engine.impl.db=trace
 #logging.level.org.activiti.engine.integration=trace
 #logging.level.org.springframework.retry=debug
+spring.cloud.stream.dynamic-destination-cache-size=50
```
igdianov, line 103: Why is it needed in the test scope? Let's remove it if not needed..

igdianov (approved):
See my comment about changed test properties

activiti-cloud-runtime-bundle-service/activiti-cloud-starter-runtime-bundle/src/test/resources/application-test.properties
astrologic7: it's needed because we have some tests that assert the `BindingProperty` object which is currently evicted by SCS...i think we can remove this with the next SCS release.

erdemedeiros (approved):

